PR #3, opened 2019-11-15
Added FASHandoverGym

m-samiec:
Create FASHandoverGym which validates GAS Effects and Abilities are handed-over correctly.


samiwh:
Can you add a user manual entry for the gym please?

PR #4, opened 2019-11-20
[UNR-2395] Add TestSuite as a TestSuite_Gym

mironec:
Adds the entirety of the old TestSuite as a gym.

Will probably need to go over it and remove old/unused code, and do a style pass.


samiwh:
Is there a list of what this tests and what we want to get from this?

mironec:
Not really at the moment, I was planning to return to this after our testing framework is finished, but I will write a quick summary here in the meantime:
- Tests replication (press K) - ints, bools, Carrays, Tarrays, RPCs, Unreal object references, etc.
- Has some simple GAS tests (press L) - adding effects, removing - mainly to test fast array serialisation
- Dormancy tests (wholly manual) - a set of 16 cubes with different behaviours when coming close to them (e.g. tear off on becoming dormant), this has to be run on native and spatial and compared.

I can add screenshots later, and I think some assets can definitely be removed (I think this has an additional version of the Player Character which is not really needed).

oblm:
What's happening with this PR?

mironec:
:( We will have to split it into two parts and clean it up, closing for now

PR #5, opened 2019-11-22
Map to test scenario where unresolved reference are replicated and progressively resolved

ImprobableNic:
Reproduction of the scenario from UNR-1132, where an array of references to actors where replicated early, when the entities were not created yet.

The level blueprint fills a replicated array of references to map actors. A RepNotify is called on clients, logging if the array is completely filled or not. Eventually, all client workers should have all references to the actors.
NB : When using zoning, some server workers won't be able to see all actors, which is actually a feature, and highlights the shortcoming of replicating Actor references.
image


joshuahuburn:
Could we get a screenshot of the map and what it looks like when working correctly.
Also, does this have a 2D UI widget in-game explaining what the test does and what to expect (so QA can run this).

mattyoung-improbable:
Could you have some info to the https://github.com/spatialos/UnrealGDKTestGyms/blob/master/USER_MANUAL.md for this new gym?

USER_MANUAL.md
```diff
@@ -55,5 +55,13 @@ Actors used in gyms are in `Contetnt\Actors`: add any new Actors to this directo
   2. Additionally a handover value is incremented to monitor how many times this is called.
   3. The stack count, and the handover counter are then checked they are the same.
 
+##### Unresolved reference gym
+* Test what happens when references to actors whose entity have not been created yet are replicated. Replicating null references is accepted, but they should be resolved eventually.
```
m-samiec:
```suggestion
* Test what happens when structs with references to actors whose entity have not been created yet are replicated. Replicating null references is accepted, but they should be resolved eventually.
```

PR #12, opened 2019-12-10
Made the handover gym hand stuff over.

samiwh:
Added more cubes, a debugger, and a 2x2 grid load balancing strategy.
Also updated the user manual for the handover gym.


USER_MANUAL.md
```diff
@@ -44,7 +44,14 @@ Actors used in gyms are in `Contetnt\Actors`: add any new Actors to this directo
 
 ##### Handover gym
 * Demonstrates that an entity can cross from one area of authority to another.
-* Contains a cube that moves back and forth across a floor.
+* Contains a set of cube that moves back and forth across a floor.
```
jennifer-at-improbable-io: nit - "a set of cubes" or "four cubes" that "move".

samiwh: will do

PR #18, opened 2020-01-08
[UNR-2583] Test gym for replicating soft reference and asynchronously loaded assets

ImprobableNic:
SoftMatRefTestGym

Test gym reproducing the issues from UNR-2583.
There was an issue with us not serializing soft references to object not loaded on the server, and another issue with replicating references to object not loaded on the client when network asynchronous loading is enabled.
This test setup cubes which takes their material from replicated properties coming from soft references on the server. I test Soft and Hard references in properties/property arrays (the array case has an additional RepNotify with a null reference).
All cubes should turn green after starting a client/server setup (although i stays red on a listen server).

NB : This test Gym turns on the net.AllowAsyncLoading CVar. I wonder how I could restore it in case we want to run multiple tests in a single session.
Also, for testing references to not loaded packages on client, the client and server should not run in the same process.


aleximprobable:
As I understand, these files `GDKTestGymsBlueprintLibrary.h/.cpp` are not useful outside of this test. Maybe we should move them to a new folder (e.g. `Game\Source\GDKTestGyms\SoftReferenceTets`)?

ImprobableNic:
@joshuahuburn  The good one ™ (4.22)
@aleximprobable These functions are specific to this one test for now, but I would rather have utility functions in a generic location, so anyone can benefit from them.

Game/Source/GDKTestGyms/GDKTestGymsBlueprintLibrary.cpp
```diff
@@ -0,0 +1,13 @@
...
+#include "Engine/Console.h"
+
+void UGDKTestGymsBlueprintLibrary::SetNetAllowAsyncLoad(bool bAllow)
+{
+  static const auto CVar = IConsoleManager::Get().FindConsoleVariable(TEXT("net.AllowAsyncLoading"));
+  if (CVar)
+  {
+    CVar->Set(bAllow);
+  }
+}
```
aleximprobable:
```suggestion
}

```

PR #21, opened 2020-01-16
Configure HandoverGym to use 4-worker grid.

yumnuska:
Depends on https://github.com/spatialos/UnrealGDK/pull/1690


yumnuska:
HandoverGymChange adds this config to the map: 
image

Game/Config/DefaultSpatialGDKSettings.ini
```diff
@@ -32,10 +32,8 @@ ActorGroups=()
 ServerWorkerTypes=("UnrealWorker")
 WorkerLogLevel=Warning
 bEnableUnrealLoadBalancer=False
```
yumnuska: Maybe load balancing should be on by default in TestGyms now?

jennifer-at-improbable-io: I think it should be on by default now. That will make it much easier to test regressions.

PR #27, opened 2020-01-28
Add GameplayAbility Locking Gym

KeyMaster-:
The ability locking gym demonstrates locking while a gameplay ability is running on an actor.
Currenty, only a server-only ability is tested - the other execution types that involve execution on the server (server initiated, local predicted) should be added in the future.

The locking will work once the corresponding PRs (GDK https://github.com/spatialos/UnrealGDK/pull/1730, Engine https://github.com/improbableio/UnrealEngine/pull/333) have been merged.

Drive-by to update the user manual for the handover gym.


jennifer-at-improbable-io:
Would we rather add this as an option to the handover gym? I'm worried about creating a new gym for each feature we add. Then again, GAS is a big enough feature we could do a gym for it. Thoughts?

KeyMaster-:
This gym should be expanded a lot more in the future with all the types of abilities. However maybe the handover test is small enough that it can be considered to be included in this gym? I.e. if anything I'd keep this gym and get rid of handover.

alastairdglennie:
I think both are useful. HandoverGym tests basic load balancing and locking in it's most exposed form. Ability system gym is a level of abstraction above this and will likely hopefully be extended into more GAS systems. Wouldn't block a merge though.

PR #30, opened 2020-02-06
[UNR-2643] Test gym for net owner relevancy

ImprobableNic:
NetOwner

This gym tests the flag to use net owner relevancy.
The gym contains two cubes, one big in the distance, with a large NetCullDistance to make it visible from the client's viewpoint, and a second smaller one with a smaller cull distance, that won't be visible, but with the bUseNetOwnerRelevancy flag set.
Every 5 seconds, the smaller one attaches itself to the bigger one and set it as its net owner.

(also contains a driveby addition of a manual entry for the soft references test gym since it seems it did not make it through on the original PR).


m-samiec:
This is great, any thoughts on a more explicit success/failure message for easier user validation?

PR #32, opened 2020-02-11
World composition gym

samiwh:
A world composition gym.

Move towards each circle to load a level.
Move away from the circles to unload the level.

image


m-samiec:
Might be worth exercising additional level types here? Add Actor's with bNetLoadOnClient set to false. Add a dormant Actor. Add an actor with a component. Thoughts?

samiwh:
Seems like a good idea. Can do.

samiwh:
What if we did something like this
image

samiwh:
Otherwise I'd suggest spending a bit more time on it and adding several levels that contain different things you can walk up to. Which seems like a good thing to have anyway.

samiwh:
Changed to a series of levels.

PR #34, opened 2020-02-11
Bugfix/double npc spawns

improbable-danny:
Fix for double spawning of NPCs which is going to impact the NFR tests

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
@@ -73,7 +73,7 @@ void ABenchmarkGymGameMode::CheckInitCustomSpawning()
 
 bool ABenchmarkGymGameMode::ShouldUseCustomSpawning()
 {
-	FString WorkerValue;		return FParse::Param(FCommandLine::Get(), TEXT("OverrideSpawning"));
+	FString WorkerValue;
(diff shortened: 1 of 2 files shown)


Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -245,19 +245,6 @@ void ABenchmarkGymGameMode::SpawnNPC(const FVector& SpawnLocation)
 
 void ABenchmarkGymGameMode::StartPlay()
 {
-	UE_LOG(LogBenchmarkGym, Log, TEXT("Enabling custom density spawning."));
```
m-samiec: How are NPCs spawned now?

improbable-danny: When CheckInitCustomSpawning is called by FindPlayerStart_Implementation, the problem with having it here is that the NetDriver/WorkerFlags class is not initialized yet!

improbable-danny: It was also double-spawning NPCs here because CheckInitCustomSpawning will also do that (it has a bool latch to ensure it only runs once)

m-samiec: Okay - perhaps just remove this `StartPlay` funciton entirely then?

PR #37, opened 2020-02-19
Fix world size and increased ground plane size in BenchmarkGym

improbable-danny:
Fix world size for spawning players in tests. Increased ground plane size + increased spatial config size


ImprobableVlad:
JSON changes look good to me. What are the changes in `BenchmarkGym.umap`?

PR #40, opened 2020-02-24
Add fps output, used by NFR pipeline

m-samiec:
FPS output is required by NFR pipeline to deduce the health of workers.

### DIFF
--- a/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.h
+++ b/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.h
@@ -19,6 +19,6 @@ private:
 	FTickerDelegate TickDelegate;
 	FDelegateHandle TickDelegateHandle;
 
-	float AverageFPS = 1.0f;
+	float AverageFPS = 60.0f;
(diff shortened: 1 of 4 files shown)


LaunchSimClient.bat
```diff
@@ -0,0 +1,4 @@
+@echo off
```
m-samiec: This is a driveby utility script

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.h
```diff
@@ -0,0 +1,24 @@
...
+
+public:
+	virtual void Init() override;
+	bool Tick(float DeltaSeconds);
+
+private:
+	FTickerDelegate TickDelegate;
+	FDelegateHandle TickDelegateHandle;
+
+	float AverageFPS = 1.0f;
```
improbable-danny: Is 30/60 a better starting point?

improbable-danny: Maybe higher is better if were going to fail tests on < x fps, the bias is removed after 30 frames from the code above

PR #41, opened 2020-02-25
Add SimulatedGameplayTask gym

improbable-valentyn:
A gym to test that simulated GameplayTasks work correctly.

`GASPawn` is a simple floating pawn (doesn't change position on the server) with an ability system component added. The component is initialized before the pawn is possessed by the controller, so I override `PossessedBy` and `OnRep_Controller` to call `RefreshAbilityActorInfo` which lets us trigger abilities from the owning client.
`SimulatedGameplayTaskPawn` is a blueprint class that extends `GASPawn`. This is the class that the pawn override is set to in `SimulatedGameplayTaskGameMode`. On the auth server, it will apply an effect to itself that will give it an ability. On the owning client, it will wait briefly to let the ability get replicated to it, and then try to activate the ability. The ability will spawn a simulated task that will be replicated to other clients.

Engine PRs:
https://github.com/improbableio/UnrealEngine/pull/367
https://github.com/improbableio/UnrealEngine/pull/368


Game/Source/GDKTestGyms/Private/Task_DelaySimulated.cpp
```diff
@@ -0,0 +1,67 @@
...
+	bSimulatedTask = true;
+}
+
+void UTask_DelaySimulated::Activate()
+{
+	UWorld* World = GetWorld();
+	FTimerHandle TimerHandle;
+	World->GetTimerManager().SetTimer(TimerHandle, this, &UTask_DelaySimulated::OnFinish, 1.0f, false);
+
+	if (USpatialGameInstance* GameInstance = Cast<USpatialGameInstance>(World->GetGameInstance()))
```
m-samiec: What's this for?

Game/Source/GDKTestGyms/Public/Task_DelaySimulated.h
```diff
@@ -0,0 +1,27 @@
...
+
+public:
+	UTask_DelaySimulated(const FObjectInitializer& ObjectInitializer);
+
+	virtual void Activate() override;
+
+	virtual void InitSimulatedTask(UGameplayTasksComponent& InGameplayTasksComponent) override;
+
+	UFUNCTION(BlueprintCallable, Category = "SimulatedGameplayTask", meta = (BlueprintInternalUseOnly = "TRUE"))
+	static UTask_DelaySimulated* TaskDelaySimulated(TScriptInterface<IGameplayTaskOwnerInterface> TaskOwner, const uint8 Priority = 192);
```
m-samiec: Magic number?

improbable-valentyn: Wanted to put `FGameplayTasks::DefaultPriority` but UHT doesn't know what it is and freaks out, so this has to be a literal.

Game/Source/GDKTestGyms/Private/Task_DelaySimulated.cpp
improbable-valentyn: Ah, I was previously printing a log with `UE_LOG` so I needed it to get the WorkerId, I'll remove this now.

Game/Source/GDKTestGyms/Public/Task_DelaySimulated.h
m-samiec: Cool - can you add a comment about what it refers to?

PR #45, opened 2020-02-28
Feature/add interest border

jacquese:
So, this is a PR that only touches binary files. It adds a 1000cm interest border to the grid LB strategy, and makes the cross server cube no longer always relevant. 
The cross server RPC test gym still works fine, and doesn't if the interest border is not added.

Happy to show the changes in person, not sure the best way to get this kind of change reviewed.


jacquese:
image
cross server cube now not always relevant
image
interest border added

PR #46, opened 2020-03-02
Spawn NPCs slower

improbable-danny:
Spawning NPCs 1 at a time to avoid a huge frame spike


Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -77,6 +79,18 @@ void ABenchmarkGymGameMode::CheckInitCustomSpawning()
 	bInitializedCustomSpawnParameters = true;
 }
 
+void ABenchmarkGymGameMode::Tick(float DeltaSeconds)
+{
+	Super::Tick(DeltaSeconds);
+	if (NPCSToSpawn > 0)
```
m-samiec: Add a check that this is only done on the auth server

PR #52, opened 2020-03-10
Added DynamicComponents test gym, verbose debugger, user_manual update

alastairdglennie:
Used for testing this GDK PR: https://github.com/spatialos/UnrealGDK/pull/1881

What it looks like:

dynamic-components

TestDynamicComponent (inherits direct from ActorComponent, it replicates, prints on AuthorityGained and has an event exposed to which our controller calls to dynamically remove itself):

image

DynamicComponentsGameMode (just uses our new player controller):

image

DynamicComponentsPlayerController (keybinds to add and remove components, stores added component added as a local ref so it can be deleted in future):

image

QuadrantLBStrategy:

image

VerboseSpatialDebugger:

image


USER_MANUAL.md
```diff
@@ -42,16 +42,36 @@ Actors used in gyms are in `Contetnt\Actors`: add any new Actors to this directo
...
+* Demonstrates that:
+  * Dynamic component are correctly added to and removed from a replicated Actor (with load-balancing enabled).
+* Contains:
+  * A character with a PlayerController with key bindings for adding and removing a component.
+* To setup the gym:
+  * Adjust the settings in "SpatialOS Settings -> Editor Settings -> Launch -> Launch configuration file options -> Server Workers" to include 4 servers in a 2x2 grid.
+    * Set both "Rectangle grid column count" and "Rectangle grid row count" to 2
+    * Set "Instances to launch in editor" to 4
+* To test the gym:
+    * Move the character until the inspector shows that the EntityACL and other server simulated components are on different workers
```
jacquese: Is there no way to set this up so that it is forced? May be beyond the scope of this PR (or should be an engine net test)

USER_MANUAL.md
```diff
@@ -42,16 +42,36 @@ Actors used in gyms are in `Contetnt\Actors`: add any new Actors to this directo
...
+    * Press "R" to add the TestDynamicComponent to the character.
+    * Observe the TestDynamicComponent added to the character components list in the inspector (and on the EntityACL)
+    * Press "T" to remove the TestDynamicComponent from the character.
```
jacquese: Do you need to do this? As a reader this feels part of the testing process, but I am not sure what I am looking for after it's happened - what constitutes a correct removal? Is it just seeing it disappear locally?

alastairdglennie: Component is removed from component list, also disappears from ComponentPresence list - will make more explicit

USER_MANUAL.md
alastairdglennie: Not afaik depends on the runtime load-balancing the EntityACL, not sure it's worth looking into fixing really, since it takes 10 seconds to run between workers and find the right setup

alastairdglennie:
Updated to use GetControlledPawn

PR #57, opened 2020-03-20
Bugfix/native bandwidth cap

improbable-danny:
Removing bandwidth cap for gyms to be a fair comparison with currently how Spatial works


Game/Config/DefaultEngine.ini
```diff
@@ -89,4 +89,12 @@ PhysXTreeRebuildRate=10
 DefaultBroadphaseSettings=(bUseMBPOnClient=False,bUseMBPOnServer=False,bUseMBPOuterBounds=False,MBPBounds=(Min=(X=0.000000,Y=0.000000,Z=0.000000),Max=(X=0.000000,Y=0.000000,Z=0.000000),IsValid=0),MBPOuterBounds=(Min=(X=0.000000,Y=0.000000,Z=0.000000),Max=(X=0.000000,Y=0.000000,Z=0.000000),IsValid=0),MBPNumSubdivs=2)
 ChaosSettings=(DefaultThreadingModel=DedicatedThread,DedicatedThreadTickMode=VariableCappedWithTarget,DedicatedThreadBufferMode=Double)
 
+; For NFR tests, we should adjust these once we have Spatial limits. This is to make a fair comparison between a playable game (without lag + banding)
+[/Script/OnlineSubsystemUtils.IpNetDriver]
+bNoTimeouts=true
```
m-samiec: This is a lot of values - you sure we need them all?

improbable-danny: I believe these are server settings and the ones below are for clients

m-samiec: To rate-limit RPCs?

ImprobableVlad:
Have you tested these changes? What numbers do they give? I'm curious about the impact, but it would also validate that these changes indeed take effect.

Also, on a kind of related note, do game developers generally stay below the default limit? I'm wondering whether the necessity for this is due to its contrived nature. If real games actually don't show this kind of activity, would it instead not be appropriate to adjust simplayer behaviour?

Game/Config/DefaultEngine.ini
improbable-danny: Yeah, those numbers are distributed to players when they join a server

improbable-danny:
> Have you tested these changes? What numbers do they give? I'm curious about the impact, but it would also validate that these changes indeed take effect.
> 
> Also, on a kind of related note, do game developers generally stay below the default limit? I'm wondering whether the necessity for this is due to its contrived nature. If real games actually don't show this kind of activity, would it instead not be appropriate to adjust simplayer behaviour?

@ImprobableVlad 

The egress rises slightly (about 20%). Without these changes there is teleporting and loads of lag (some simplayers were completely starved and had no activity). Games probably wouldn't uncap this like we have but we know we have plenty of bandwidth available on the server - really we want to know how much bandwidth is used to have a stable playable experience with our gym (comparable to Spatial). In the future we will bring these numbers back down and implement a similar feature for Spatial. 

I suspect any medium/large scale game will hit the defaults so most probably will need to tweak it at some point, there are few options to deal with it such as the replication graph or using the dynamic GameNetworkManager which will adjust the available bandwidth to clients. Perhaps we should try this at some point also as well

Game/Config/DefaultEngine.ini
```diff
@@ -89,4 +89,12 @@ PhysXTreeRebuildRate=10
 DefaultBroadphaseSettings=(bUseMBPOnClient=False,bUseMBPOnServer=False,bUseMBPOuterBounds=False,MBPBounds=(Min=(X=0.000000,Y=0.000000,Z=0.000000),Max=(X=0.000000,Y=0.000000,Z=0.000000),IsValid=0),MBPOuterBounds=(Min=(X=0.000000,Y=0.000000,Z=0.000000),Max=(X=0.000000,Y=0.000000,Z=0.000000),IsValid=0),MBPNumSubdivs=2)
 ChaosSettings=(DefaultThreadingModel=DedicatedThread,DedicatedThreadTickMode=VariableCappedWithTarget,DedicatedThreadBufferMode=Double)
 
+; For NFR tests, we should adjust these once we have Spatial limits. This is to make a fair comparison between a playable game (without lag + banding)
```
ImprobableVlad:
```suggestion
; For NFR tests. This is to make a fair comparison between GDK and Native Unreal for a playable game (without lag + banding)
; TODO: We should adjust these once we have Spatial limits.
```

PR #61, opened 2020-03-26
Add cube destroy bind for AbilityLockingGym

alastairdglennie:
Will add screenies

AbilityLockingGym:

image

AbilityLockingController:

image

Counting Cube:

image


m-samiec:
Why do we need to trigger an ability on the character?

PR #62, opened 2020-03-27
Bugfix/client fps

improbable-danny:
Fixes for 4.24

- Server FPS set to 60, clients 30 (possibly a 4.24 bug introduced with client fps with nullrhi)
- Renaming gyms used by nfr pipeline because they only allow lowercase names atm (urgent fix)


Game/Config/DefaultEngine.ini
```diff
@@ -25,6 +25,8 @@ AppliedDefaultGraphicsPerformance=Maximum
 +ActiveClassRedirects=(OldClassName="StarterProjectGameMode",NewClassName="GDKTestGymsGameMode")
 +ActiveClassRedirects=(OldClassName="StarterProjectCharacter",NewClassName="GDKTestGymsCharacter")
 WorldSettingsClassName=/Script/SpatialGDK.SpatialWorldSettings
+bUseFixedFrameRate=True
+FixedFrameRate=60.0
```
m-samiec: Add a comment here that we override this in servers to run at 30

Game/Source/GDKTestGyms/NFRGameMode.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+void ANFRGameMode::StartPlay()
+{
+	if (GetWorld()->IsServer())
+	{
+		// Server only. This is a because of 4.24 bug where client FPS cannot be set with nullrhi.
+		// Clients are set to 60fps by DefaultEngine settings bUseFixedFrameRate and FixedFrameRate.
+		// Hopefully both of these will future-proof us against this breaking, only FixedFrameRate is used in 4.24 but in 4.23 MaxFPS is used.
+		GEngine->SetMaxFPS(30.0f); 
+		GEngine->FixedFrameRate = 30.0f;
```
m-samiec: Is this available in 4.23? If not add a version #def

improbable-danny: Yeah compiles with 4.23 also

PR #65, opened 2020-04-01
Feature/GV-255 Concurrent latency tests with same assembly

ImprobableVlad:
This is the sister PR to [this PR](https://github.com/improbable/TestGymBuildKite/pull/54), which gives more context for these changes are.

The whole commented block has been removed from the LatencyGym map blueprint. It contained (convoluted) logic for extracing a command line argument, which is now built into the GDK after a recent PR.
Annotation 2020-03-24 175540

**Testing**
A test build has been kicked off here:
https://buildkite.com/improbable/unrealgdk-nfr/builds/1256
If I'm still around at that point I will cancel all non-latency tests.


spatial/workers/unreal/spatialos.UnrealWorker.worker.json
```diff
@@ -76,9 +74,13 @@
         "${IMPROBABLE_WORKER_NAME}",
         "+workerId",
         "${IMPROBABLE_WORKER_ID}",
+		"+deploymentName",
```
m-samiec, line 77: spacing?

ImprobableVlad: yup, fixed it now!

PR #66, opened 2020-04-03
[UNR-3206] Fix NetReferenceTestGym to manually spawn controller after authority has been gained

ImprobableNic:
Was broken because of https://improbableio.atlassian.net/browse/UNR-3231

Fixed it to not use this code path anymore and explicitely spawn a player controller when authority has been gained.
CheckoutPawn
CheckoutPawnController


joshuahuburn:
Can we have a couple screenshots of the BPs?

m-samiec:
Why is the "delay" necessary?

ImprobableNic:
Can't connect the output pin to the input pin otherwise.

m-samiec:
> Can't connect the output pin to the input pin otherwise.

Really? There's an increment inbetween and it's an async action, so I don't see what the problem would be. Whats the error?

ImprobableNic:
> Really? There's an increment inbetween and it's an async action, so I don't see what the problem would be. Whats the error?

IIRC, it just crashes at runtime with a blueprint VM infinite loop error.

PR #72, opened 2020-04-09
[UNR-3209][MS] Adding a gym to test possession and sending RPCs on multiple player controlled actors actors.

MatthewSandfordImprobable:
image

Hitting "space" will cause a possession of the other cube in the gym. It will also set the owner of the previously unpossesed pawn to the player controller.

Hitting "enter" will print out information on the client and also attempt to send server RPCs on all cubes. The implementation of these RPCs simply prints out some text.  

Cube pawn:
image

Player Controller:
image
image


aleximprobable:
A few nits:
1. Is it necessary to have a `Break` condition in `For Each` loop in `Controller`?
2. Is it worth putting the `Possess New Cube` blueprint part into a separate Function in `Controller`?

MatthewSandfordImprobable:
I think I will keep the break in the loop. If we added more cubes and did not have the break we would have unwanted behavior. 

As for adding a function, good idea. I usually do do this just was lazy I guess :p

PR #74, opened 2020-04-23
Updated settings

improbable-danny:
Updating settings for lowest latency at a small bandwidth cost (<2%)


improbable-danny:
We have decided to not do this just yet

PR #92, opened 2020-05-27
Updating FPS metric to match GDK

improbable-danny:
Updating the FPS metric to use a 2 minute window similar to the GDK metric

### DIFF
--- a/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
+++ b/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
@@ -10,6 +10,7 @@ void UGDKTestGymsGameInstance::Init()
 
 	TickDelegate = FTickerDelegate::CreateUObject(this, &UGDKTestGymsGameInstance::Tick);
 	TickDelegateHandle = FTicker::GetCoreTicker().AddTicker(TickDelegate);
+	TickWindowTotal = 0;
 }
 
 void UGDKTestGymsGameInstance::OnStart()
@@ -23,26 +24,34 @@ void UGDKTestGymsGameInstance::OnStart()
 
 float UGDKTestGymsGameInstance::AddAndCalcFps(int64 NowReal, float DeltaS)
 {
-	TicksForFPS.Push(TPair<int64, float>(NowReal, DeltaS)); 
-	float TotalDelta = 0.0f;
-	int NumSamples = 0;
+	int64 DeltaTicks = FTimespan::FromSeconds(DeltaS).GetTicks();
+	TicksForFPS.Push(TPair<int64, int64>(NowReal, DeltaTicks)); 
+	TickWindowTotal += DeltaTicks;
+
 	int64 MinRange = NowReal - FTimespan::FromMinutes(2.0).GetTicks();
-	for (int i = TicksForFPS.Num()-1; i >= 0; i--)
+	
+	// Trim samples
+	for(int i = 0; i < TicksForFPS.Num(); i++)
 	{
 		const FPSTimePoint& TimePoint = TicksForFPS[i];
-		if (TimePoint.Key < MinRange)
+		if (TimePoint.Key > MinRange) // Find the first valid sample
 		{
-			TicksForFPS.RemoveAt(i);
-		}
-		else
-		{
-			TotalDelta += TimePoint.Value;
-			NumSamples++;
+			int NumToRemove = i;
+			if (NumToRemove > 0)
+			{
+				for (int j = 0; j < NumToRemove; j++)
+				{
+					TickWindowTotal -= TicksForFPS[j].Value;
+				}
+				memmove(&TicksForFPS[0], &TicksForFPS[NumToRemove], (TicksForFPS.Num()-NumToRemove) * sizeof(FPSTimePoint));
+				TicksForFPS.SetNum(TicksForFPS.Num() - NumToRemove);
+			}
+			break;
 		}
 	}
-	if (NumSamples > 0)
+	if (TicksForFPS.Num() > 0)
 	{
-		return static_cast<float>(NumSamples) / TotalDelta; // 1.0f / Total / Samples 
+		return static_cast<float>(TicksForFPS.Num()) / FTimespan(TickWindowTotal).GetTotalSeconds(); // 1.0f / Total / Samples 
 	}
 	return 60.0f; // If there are no samples return the ideal
 }
--- a/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
+++ b/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
@@ -37,7 +37,7 @@ float UGDKTestGymsGameInstance::AddAndCalcFps(int64 NowReal, float DeltaS)
 		const FPSTimePoint& TimePoint = TicksForFPS[i];
 		if (TimePoint.Key > MinRange) // Find the first valid sample
 		{
-			int NumToRemove = i;
+			NumToRemove = i;
 			break;
 		}
 	}
--- a/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
+++ b/Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
@@ -43,9 +43,9 @@ float UGDKTestGymsGameInstance::AddAndCalcFps(int64 NowReal, float DeltaS)
 	}
 	if (NumToRemove > 0)
 	{
-		for (int j = 0; j < NumToRemove; j++)
+		for (int i = 0; i < NumToRemove; i++)
 		{
-			TickWindowTotal -= TicksForFPS[j].Value;
+			TickWindowTotal -= TicksForFPS[i].Value;
 		}
 		memmove(&TicksForFPS[0], &TicksForFPS[NumToRemove], (TicksForFPS.Num() - NumToRemove) * sizeof(FPSTimePoint));
 		TicksForFPS.SetNum(TicksForFPS.Num() - NumToRemove);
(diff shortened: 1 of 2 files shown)


Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +21,42 @@ void UGDKTestGymsGameInstance::OnStart()
...
+	TicksForFPS.Push(TPair<int64, float>(NowReal, DeltaS)); 
+	float TotalDelta = 0.0f;
+	int NumSamples = 0;
+	int64 MinRange = NowReal - FTimespan::FromMinutes(2.0).GetTicks();
+	for (int i = TicksForFPS.Num()-1; i >= 0; i--)
+	{
+		const FPSTimePoint& TimePoint = TicksForFPS[i];
+		if (TimePoint.Key < MinRange)
+		{
+			TicksForFPS.RemoveAt(i);
```
m-samiec, line 35: order doesn't seem important, so can use RemoveAtSwap

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +21,42 @@ void UGDKTestGymsGameInstance::OnStart()
...
+		{
+			TotalDelta += TimePoint.Value;
+			NumSamples++;
+		}
+	}
+	if (NumSamples > 0)
+	{
+		return static_cast<float>(NumSamples) / TotalDelta; // 1.0f / Total / Samples 
+	}
+	return 60.0f; // If there are no samples return the ideal
```
m-samiec, line 47: This isn't possible right?

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +21,42 @@ void UGDKTestGymsGameInstance::OnStart()
...
+float UGDKTestGymsGameInstance::AddAndCalcFps(int64 NowReal, float DeltaS)
 {
-	float Alpha = 0.8f;
-	float CurrFPS = 1.0f / DeltaSeconds; // TODO - we should remove FApp::GetIdleTime() from DeltaSeconds once the NFR regex is fixed
-	AverageFPS = (Alpha * AverageFPS) + ((1.0f - Alpha) * CurrFPS);
+	TicksForFPS.Push(TPair<int64, float>(NowReal, DeltaS)); 
+	float TotalDelta = 0.0f;
+	int NumSamples = 0;
+	int64 MinRange = NowReal - FTimespan::FromMinutes(2.0).GetTicks();
+	for (int i = TicksForFPS.Num()-1; i >= 0; i--)
```
MatthewSandfordImprobable, line 30: If TicksForFPS is in order we could loop up until we find a good time and then break.

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +21,42 @@ void UGDKTestGymsGameInstance::OnStart()
...
+		{
+			TicksForFPS.RemoveAt(i);
+		}
+		else
+		{
+			TotalDelta += TimePoint.Value;
+			NumSamples++;
+		}
+	}
+	if (NumSamples > 0)
```
MatthewSandfordImprobable, line 43: Could we not just use TicksForFPS.Num() here. Do we need to increment NumSamples in the loop?

MatthewSandfordImprobable: Can we calculate the "TotalDelta" from the Datetime ticks? 
We could look at the first element and last element of TicksForFPS, and perhaps use their "NowReal" times to calculate the DeltaTime?

MatthewSandfordImprobable: This might also allow us to break early from the loop once we have removed all elements who fall outside the window (if TicksForFPS is ordered which I guess it is)

MatthewSandfordImprobable: If I have explained this poorly we can have a chat :)

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
improbable-danny: Probably not, unless the delta passed in fails the 2 minute check

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
improbable-danny: `Do we need to increment NumSamples in the loop?` Good point can just use the array size
`Can we calculate the "TotalDelta" from the Datetime ticks?` Good point we possibly could, although we use DeltaTime for the Spatial metrics too and this doesn't always match real time (max delta capping). It's worth thinking about performance though because the array length is ~7200 so it could be worth thinking about how we can accumulate better - maybe using the sliding window method of adding+subtracting when a value is added/removed (but without float to avoid precision loss over long periods)

MatthewSandfordImprobable: Yeah maybe having a member variable that gets added to whenever we get a new DeltaSeconds. We also subtract any TimePoints deltas that were not in the 2 min window. This allows us to only loop over a few elements and break early.

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +22,50 @@ void UGDKTestGymsGameInstance::OnStart()
...
+	{
+		const FPSTimePoint& TimePoint = TicksForFPS[i];
+		if (TimePoint.Key > MinRange) // Find the first valid sample
+		{
+			int NumToRemove = i;
+			if (NumToRemove > 0)
+			{
+				for (int j = 0; j < NumToRemove; j++)
+				{
+					TickWindowTotal -= TicksForFPS[j].Value;
```
MatthewSandfordImprobable, line 44: perhaps we could do: 

if (TimePoint.Key > MinRange) // Find the first valid sample
{
   logic
}
else
{
    TickWindowTotal -= TicksForFPS[i].Value;
}

MatthewSandfordImprobable: Nit: This is more preference but you might think this makes things easier to read.
I would have NumToRemove declared outside the loop. The loop simply sets NumToRemove and breaks.
After the loop we remove the elements from TicksForFPS. 

Feel free to go with whatever you prefer here. For me my suggestion just breaks things up into two steps and removes some indentation.

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +22,50 @@ void UGDKTestGymsGameInstance::OnStart()
...
+				}
+				memmove(&TicksForFPS[0], &TicksForFPS[NumToRemove], (TicksForFPS.Num()-NumToRemove) * sizeof(FPSTimePoint));
```
MatthewSandfordImprobable, line 46: I wonder if we can remove the elements from TicksForFPS using TArray members.
This might allow us to remove the: TicksForFPS.SetNum(TicksForFPS.Num() - NumToRemove);

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
improbable-danny: It was initially like that but it sort of combined 2 steps together which made things slightly unclear. I like the 2 steps, makes sense and breaks it up

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
improbable-danny: It does but unfortunately can only remove one at a time, if we ever need to remove more than 1 we incur the cost of moving all the array elements multiple times (at least I couldn't see an API that removes a range). Perhaps a better data structure but that has its own complexity overheads (btree rebalancing) - since we only do this once per frame it may not be worth optimizing away too much, we could rearrange it to only remove when we calculate the FPS

MatthewSandfordImprobable: Yeah, if there isn't anything nice on TArray then I am completely happy!

improbable-danny:
Test running here https://buildkite.com/improbable/unrealgdk-nfr/builds/1779

Game/Source/GDKTestGyms/GDKTestGymsGameInstance.cpp
```diff
@@ -21,19 +22,51 @@ void UGDKTestGymsGameInstance::OnStart()
...
+		const FPSTimePoint& TimePoint = TicksForFPS[i];
+		if (TimePoint.Key > MinRange) // Find the first valid sample
+		{
+			NumToRemove = i;
+			break;
+		}
+	}
+	if (NumToRemove > 0)
+	{
+		for (int j = 0; j < NumToRemove; j++)
```
MatthewSandfordImprobable, line 46: Omega Nit: the loop can use "i" now XD

MatthewSandfordImprobable: Also Omega Nit: TickWindowTotal -= TicksForFPS[j].Value logic could be performed in the previous loop. This would reduce ops but....... Also you kinda have this nice separation of functionality that would be broken.

I don't know which of these I prefer tbh I just wanted to point it out

improbable-danny: Yeah, I think it was slightly confusing when it was in the loop since the 2 branches were together and the logic for 2 things was merged

PR #94, opened 2020-05-29
Updated WorldSettings on all load balanced or offloaded test gym maps

jennifer-at-improbable-io:
Layer settings (was ActorGroups) have been moved from the Editor/Runtime settings into WorldSettings. This fixes any TestGym maps that had LB enabled.

This also adds the OffloadableGymCube asset and an Offloading Test Gym map, which takes the Handover gym with a 2x2 grid strategy and a singleton offloaded strategy.


mironec:
Offloadable gym cube is also updated?

jennifer-at-improbable-io:
OffloadableGymCube is a new asset I've added for the new Offloading Gym. I'll update the description to include that.

PR #113, opened 2020-06-12
Disabling multi worker in offloading map

MatthewSandfordImprobable:
Also increasing NCD on NPCs


MatthewSandfordImprobable:
> 90 commits!?

Yeah not sure what is going on here. 90% of the commits are already in master...

PR #129, opened 2020-06-17
[UNR-3678][MS] Consistent metric timers for GDK and native

MatthewSandfordImprobable:
This is what we want to achieve:

- In spatial we want to "fail" without any delay. This means that we will receive useful logs throughout the lifetime of a worker. We will rely on the initial delays set in NFR yaml files to ensure these metric do not fail tests too early

- In native we will not report these logs before a certain timer. This is due to the fact that the logs are used to fail NFR runs. 

All timers have been added to NFRConstants for consistency.

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -39,6 +39,7 @@ FString ABenchmarkGymGameModeBase::ReadFromCommandLineKey = TEXT("ReadFromComman
 
 ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	: ExpectedPlayers(1)
+	, bIsUsingSpatialNetworking(false)
 	, PrintUXMetric(10.0f)
 	, MaxClientRoundTripSeconds(150)
 	, MaxClientUpdateTimeDeltaSeconds(300)
@@ -63,7 +64,7 @@ void ABenchmarkGymGameModeBase::BeginPlay()
 {
 	Super::BeginPlay();
 
-	IsUsingSpatialNetworking = GetDefault<UGeneralProjectSettings>()->UsesSpatialNetworking();
+	bIsUsingSpatialNetworking = GetDefault<UGeneralProjectSettings>()->UsesSpatialNetworking();
 
 	ParsePassedValues();
 	TryBindWorkerFlagsDelegate();
@@ -98,7 +99,7 @@ void ABenchmarkGymGameModeBase::TryBindWorkerFlagsDelegate()
 
 void ABenchmarkGymGameModeBase::TryAddSpatialMetrics()
 {
-	if (!IsUsingSpatialNetworking)
+	if (!bIsUsingSpatialNetworking)
 	{
 		return;
 	}
@@ -169,11 +170,11 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
 		return;
 	}
 
-	const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
+	UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
 	check(Constants);
 
 	// This test respects the initial delay timer in both native and GDK
-	if (Constants->PlayerCheckValid())
+	if (Constants->PlayerCheckMetricDelay.IsReady())
 	{
 		bHasDonePlayerCheck = true;
 		if (ActivePlayers != ExpectedPlayers)
@@ -210,17 +211,19 @@ void ABenchmarkGymGameModeBase::TickServerFPSCheck(float DeltaSeconds)
 		return;
 	}
 
-	const UNFRConstants* Constants = UNFRConstants::Get(World);
+	UNFRConstants* Constants = UNFRConstants::Get(World);
 	check(Constants);
 
 	const float FPS = GameInstance->GetAveragedFPS();
 	if (FPS < Constants->GetMinServerFPS())
 	{
-		if (IsUsingSpatialNetworking)
+		// Immediately fail for SpatialOS as we do not need any initial delay. The NFR metrics will use their internal delay.
+		// We still want to print logs in the case of SpatialOS.
+		if (bIsUsingSpatialNetworking)
 		{
 			FailServerFPSCheck(FPS);
 		}
-		else if (Constants->SamplesForServerFPSValid())
+		else if (Constants->ServerFPSMetricDelay.IsReady())
 		{
 			FailServerFPSCheck(FPS);
 		}
@@ -258,16 +261,16 @@ void ABenchmarkGymGameModeBase::TickClientFPSCheck(float DeltaSeconds)
 
 	if (!bClientFpsWasValid)
 	{
-		if (IsUsingSpatialNetworking)
+		if (bIsUsingSpatialNetworking)
 		{
 			FailClientFPSCheck();
 		}
 		else
 		{
-			const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
+			UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
 			check(Constants);
 
-			if (Constants->SamplesForServerFPSValid())
+			if (Constants->ServerFPSMetricDelay.IsReady())
 			{
 				FailClientFPSCheck();
 			}
@@ -335,16 +338,16 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
 
 	if (AveragedClientRTTSeconds > MaxClientRoundTripSeconds || AveragedClientUpdateTimeDeltaSeconds > MaxClientUpdateTimeDeltaSeconds)
 	{
-		if (IsUsingSpatialNetworking)
+		if (bIsUsingSpatialNetworking)
 		{
 			FailUXMetricCheck();
 		}
 		else
 		{
-			const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
+			UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
 			check(Constants);
 
-			if (Constants->UXMetricValid())
+			if (Constants->UXMetricDelay.IsReady())
 			{
 				FailUXMetricCheck();
 			}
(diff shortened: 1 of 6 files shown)


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -42,16 +42,17 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 
 private:
 
+	bool IsUsingSpatialNetworking;
```
improbable-danny, line 45:
```suggestion
	bool bIsUsingSpatialNetworking;
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -39,13 +39,12 @@ FString ABenchmarkGymGameModeBase::ReadFromCommandLineKey = TEXT("ReadFromComman
...
 ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	: ExpectedPlayers(1)
-	, SecondsTillPlayerCheck(15.0f * 60.0f)
 	, PrintUXMetric(10.0f)
 	, MaxClientRoundTripSeconds(150)
 	, MaxClientUpdateTimeDeltaSeconds(300)
-	, bPlayersHaveJoined(false)
 	, bHasUxFailed(false)
 	, bHasFpsFailed(false)
+	, bHasDonePlayerCheck(false)
```
improbable-danny, line 47: Init bIsUsingSpatialNetworking here too

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -162,50 +163,89 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
...
+
+	const UNFRConstants* Constants = UNFRConstants::Get(World);
 	check(Constants);
-	if (Constants->SamplesForFPSValid() && !bHasFpsFailed && GetWorld() != nullptr)
+
+	const float FPS = GameInstance->GetAveragedFPS();
+	if (FPS < Constants->GetMinServerFPS())
 	{
-		if (const UGDKTestGymsGameInstance* GameInstance = Cast<UGDKTestGymsGameInstance>(GetWorld()->GetGameInstance()))
+		if (IsUsingSpatialNetworking)
```
improbable-danny, line 219: Could be worth adding a comment here about why we immediately fail for spatial

improbable-danny: Also I don't think this behavior matches now we use the boolean fps check

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
```diff
@@ -7,24 +7,48 @@
 
 DEFINE_LOG_CATEGORY(LogNFRConstants);
 
-bool UNFRConstants::SamplesForFPSValid() const
+bool UNFRConstants::DoValidityCheck(bool& IsValid, int64 TimeToStart) const
```
improbable-danny, line 10: Nit:
What if we wrap the timing code (and time that it becomes valid) up in a little struct which we can call .IsValid which internally does the caching behavior rather than passing references to a bool here

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
improbable-danny: Think wrapping in `SamplesForServerFPSValid` should fix it, we want to upload 1 until we start actually checking

MatthewSandfordImprobable: bHasFpsFailed is false by default so GetFPSValid will return 1 by default. I think this matches?

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
MatthewSandfordImprobable: Sure sounds good.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
improbable-danny: Oops yeah, might still be slightly incorrect though - we can fail from the start of the deployment which doesn't match native. It's slightly different for this because we use the metric as a signal instead of an actual value so the 2 minute window is on the signal not the FPS check (for which the signal only goes from 1->0 which means the deployment has failed)

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -162,50 +161,73 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
 		return;
 	}
 
-	if (SecondsTillPlayerCheck > 0.0f)
+	// Only check players once
+	if (bHasDonePlayerCheck)
```
m-samiec, line 165: We also want to catch players disconnecting during the game, so I don't think we should have this flag.

m-samiec: Oh - we didn't do that previously - that seems incorrect.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -274,17 +295,26 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
+		!bUXMetricValid &&
+		Constants->UXMetricDelay.IsReady())
 	{
 		bHasUxFailed = true;
 		NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("UX metric has failed. RTT: %.8f, UpdateDelta: %.8f, ActivePlayers: %d"), AveragedClientRTTSeconds, AveragedClientUpdateTimeDeltaSeconds, ActivePlayers);
 	}
-
-	PrintUXMetric -= DeltaSeconds;
-	if (PrintUXMetric < 0.0f)
+	else
```
m-samiec, line 310: Why else?

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
```diff
@@ -7,25 +7,32 @@
...
 
-void UNFRConstants::InitWithWorld(UWorld* World)
+bool FMetricDelay::IsReady() const
 {
-	TimeToStartFPSSampling = FDateTime::Now().GetTicks() + FTimespan::FromSeconds(120.0f).GetTicks();
+	return FDateTime::Now().GetTicks() > TimeToStart;
+}
 
+UNFRConstants::UNFRConstants()
+	: PlayerCheckMetricDelay(FDateTime::Now().GetTicks() + FTimespan::FromSeconds(15.0f * 60.0f).GetTicks())
```
m-samiec, line 27: Just pass in the delta, and let the Delay object calculate the offset.

Game/Source/GDKTestGyms/Public/NFRConstants.h
```diff
@@ -8,24 +8,44 @@
 
 DECLARE_LOG_CATEGORY_EXTERN(LogNFRConstants, Log, All);
 
+class FMetricDelay
+{
+public:
+
+	FMetricDelay();
```
m-samiec, line 15: Do we need this constructor?

Game/Source/GDKTestGyms/Public/NFRConstants.h
```diff
@@ -8,24 +8,44 @@
...
+	int64 TimeToStart;
+};
+
 UCLASS()
 class UNFRConstants : public UObject
 {
 public:	
+
+	UNFRConstants();
+
```
m-samiec, line 30: Please put generated_body at the top of the class

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
MatthewSandfordImprobable: TRU!

Game/Source/GDKTestGyms/Public/NFRConstants.h
MatthewSandfordImprobable: I get errors stating that "no appropriate default constructor available"

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
improbable-danny: The native fail log should be in GameMode::Logout too

Game/Source/GDKTestGyms/Public/NFRConstants.h
```diff
@@ -8,24 +8,46 @@
 
 DECLARE_LOG_CATEGORY_EXTERN(LogNFRConstants, Log, All);
 
+class FMetricDelay
+{
+public:
+
+	FMetricDelay();
```
improbable-danny, line 15:
```suggestion
	FMetricDelay() = default;
```

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
```diff
@@ -7,25 +7,35 @@
...
-	bool bIsValidNow = FDateTime::Now().GetTicks() > TimeToStartFPSSampling;
-	if (bIsValidNow)
-	{
-		bFPSSamplingValid = true;
-	}
-	return bIsValidNow;
 }
 
-void UNFRConstants::InitWithWorld(UWorld* World)
+FMetricDelay::FMetricDelay(int64 Seconds)
```
m-samiec, line 14:
```suggestion
FMetricDelay::FMetricDelay(float Seconds)
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -162,50 +161,73 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
...
-		SecondsTillPlayerCheck -= DeltaSeconds;
-		if (SecondsTillPlayerCheck <= 0.0f)
+		return;
+	}
+
+	const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
+	check(Constants);
+
+	// This test respects the initial delay timer in both native and GDK
+	if (Constants->PlayerCheckMetricDelay.IsReady())
```
improbable-danny, line 174: This is really nice

PR #131, opened 2020-06-18
UNR-3689 - Update gym user guide for new worker instance launching UI

alastairdglennie:


USER_MANUAL.md
```diff
@@ -37,6 +37,14 @@ Actors used in gyms are in `Contetnt\Actors`: add any new Actors to this directo
 4. Add a description of the new gym level to this `USER_MANUAL.md` document, in the section below;
   * breifly describe what it demonstrates and how to use it.
 
+#### Load balancing configuration
+
+Some gyms require configuring 4 server worker instances to launch. You can do this by:
+
+* Open the settings in `SpatialOS Settings` -> `Editor Settings` -> `Launch` -> `Server Worker Config`.
```
mironec, line 44: Wouldn't you need a quadrant LB strategy in the specified maps anyway?
Then you shouldn't need to set this? I am probably missing something.

alastairdglennie: I think we can and should set them all by default?

alastairdglennie: Half of them already had, just updated the others

PR #135, opened 2020-06-24
[UNR-3718][MS] Adding clarity to the multiple ownership gym

MatthewSandfordImprobable:
Fixing a blueprint error: PreviousPawn can be null if we have not previously possessed anything. I added an "IsValid" check...
image

Added a camera to the multiple ownership map and set the player controller to use the camera initially:
image


USER_MANUAL.md
```diff
@@ -216,7 +216,7 @@ Actors used in gyms are in `Contetnt\Actors`: add any new Actors to this directo
 * Demonstrates sending RPCs on multiple actors that have their owner set to a player controller. 
 * Pressing "space" will switch the possession between the two cubes in the gym. This will also ensure that the unpossessed cube will still be owned by the player controller.
 * Pressing "enter" will print out information on the client regarding the owners of each cube. It will also send a server RPC that will print out if it was successfully processed on the server. 
-* Note that initially no cube is possessed so all server RPCs sent on the cubes will fail. After a cube is possessed, the server RPC called on a cube that does not have a player controller owner will not be processed.
+* Note that initially no cube is possessed so all server RPCs sent on the cubes will fail. This means there will be no messaged printed on the server. After a cube is possessed, a server RPC called on a cube that does not have a player controller owner will not be processed.
```
mironec, line 219:
```suggestion
* Note that initially no cube is possessed so all server RPCs sent on the cubes will fail. This means there will be no messages printed on the server. After a cube is possessed, a server RPC called on a cube that does not have a player controller owner will not be processed.
```

PR #139, opened 2020-06-29
Adding metric to track the expected number of objects within a scenario

MatthewSandfordImprobable:
A single metric is used to denote whether the scenario has failed due to an unexpected number of certain ActorClasses. This will be hooked into using the NFR yaml files. 

Derived benchmark game modes can specify the actors they want to consider and the expected amount in the scenario. A variance can be specified to allow some leeway. 

To calculate the number of actors in the scenario the benchmark game mode uses the counter component on the game state. This is to reduce the number of GetAllActorsofClass calls. This means that when adding a new actor class to the keep track of on the game mode we also need to ensure the game state is tracking that actor. 

The counter component has been moved from blueprints to code so that it can be accessed from the ABenchmarkGymGameModeBase.

Linked: https://github.com/improbable/TestGymBuildKite/pull/134


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -37,12 +39,34 @@ namespace
...
+	, Timer(ResetTime)
+{}
+
+void FPrintTimer::SetResetTimer(float InResetTime)
+{
+	ResetTime = InResetTime;
+	Timer = InResetTime;
+}
+
+void FPrintTimer::Tick(float DeltaSeconds)
```
improbable-danny, line 53: What if this was TickAndCheck and returned a bool? Is this a cleaner API?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -313,14 +351,43 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
+	if (!bHasActorCountFailed &&
+		Constants->ActorCheckDelay.IsReady())
+	{
+		for (const FExpectedActorCount& ExpectedActorCount : ExpectedActorCounts)
+		{
+			const FString& ActorClassName = ExpectedActorCount.ActorClass->GetName();
+			const int32 ExpectedCount = ExpectedActorCount.ExpectedCount;
+			const int32 Variance = ExpectedActorCount.Variance;
+			const int32 ActualCount = GetActorClassCount(ExpectedActorCount.ActorClass);
+			bHasActorCountFailed = ActualCount < ExpectedCount - Variance || ActualCount > ExpectedCount + Variance;
```
improbable-danny, line 380:
```suggestion
			bHasActorCountFailed = abs(ActualCount - ExpectedCount) < Variance;
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -414,4 +481,12 @@ void ABenchmarkGymGameModeBase::SetTotalNPCs(int32 Value)
...
 {
 	OnTotalNPCsUpdated(TotalNPCs);
+}
+
+int32 ABenchmarkGymGameModeBase::GetActorClassCount(TSubclassOf<AActor> ActorClass) const
+{
+	const UCounterComponent* CounterComponent = Cast<UCounterComponent>(GameState->GetComponentByClass(UCounterComponent::StaticClass()));
+	check(CounterComponent != nullptr)
+
+	return CounterComponent->GetActorClassCount(ActorClass);
```
improbable-danny, line 491: I guess this could actually be a static function somewhere so we don't have to rely on a counter component. Perhaps both this and counter component can reference it (or maybe a static on counter component that takes a world and class?)

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -10,6 +10,25 @@
 
 DECLARE_LOG_CATEGORY_EXTERN(LogBenchmarkGymGameModeBase, Log, All);
 
+class FPrintTimer
+{
+public:
+
+	FPrintTimer() = default;
```
improbable-danny, line 17: Should we remove this and enforce initilization with a time?

Game/Source/GDKTestGyms/BenchmarkGymNPCSpawner.cpp
```diff
@@ -9,15 +9,15 @@ DEFINE_LOG_CATEGORY(LogBenchmarkGymNPCSpawner);
...
-	static ConstructorHelpers::FClassFinder<APawn> NPCBPClass(TEXT("/Game/Characters/SimulatedPlayers/BenchmarkNPC_BP"));
-	if (NPCBPClass.Class != NULL)
-	{
-		NPCPawnClass = NPCBPClass.Class;
-	}
 	SetReplicates(true);
 	bAlwaysRelevant = true;
 }
 
+void ABenchmarkGymNPCSpawner::SetNPCClass(TSubclassOf<APawn> NewNPCClass)
```
improbable-danny, line 16: Did you make this change to share the NPC spawner? If so, nice

Game/Source/GDKTestGyms/OffloadingBenchmarkGymGameMode.cpp
```diff
@@ -0,0 +1,37 @@
...
+	ConstructorHelpers::FClassFinder<AActor>  SimulatedPlayerBPClassFinder(TEXT("/Game/OffloadingBenchmark/OffloadingBenchmark_NPC_BP"));
+	SimulatedPlayerBPClass = SimulatedPlayerBPClassFinder.Class;
+}
+
+void AOffloadingBenchmarkGymGameMode::BuildExpectedObjectCounts()
+{
+	Super::BuildExpectedObjectCounts();
+
+	{
+		FExpectedActorCount ExpectedActorCount;
```
m-samiec, line 23: Could we add an check for the dorp cube - it's most likely the one that breaks currently.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -10,6 +10,25 @@
 
 DECLARE_LOG_CATEGORY_EXTERN(LogBenchmarkGymGameModeBase, Log, All);
 
+class FPrintTimer
```
m-samiec, line 13: nit - no need to prefix with F if it's not an unreal struct

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: So it could be. At one point I was checking ShouldPrint in two different places. If we implemented TickAndCheck this would mean we would have to cache the bool somewhere and then use that if we needed to print in two separate locations. 

That said I am no longer printing in two different parts of the code using this timer, however I like having that option.  
Let me know what you think.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
MatthewSandfordImprobable: Ahh yes, forgot it's not an unreal class.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: The CounterComponent will cache the actor class value. I use the game state counter component because I do not want to call too many GetAllActorsOfClass.

Game/Source/GDKTestGyms/OffloadingBenchmarkGymGameMode.cpp
MatthewSandfordImprobable: Yes we can, my plan was to hand this to Danny once my this code is in. He can then add what is required for the benchmark gym. This was for the purpose of getting someone to use my "API" and also because Danny has a better understanding of the benchmark gym and can add all the required actors we need to keep track of. If there isn't much to add I can do it in this PR, just gimme a list of actors and numbers.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -32,6 +30,9 @@ ABenchmarkGymGameMode::ABenchmarkGymGameMode()
 {
 	PrimaryActorTick.bCanEverTick = true;
 
+	ConstructorHelpers::FClassFinder<APawn> NPCBPClassFinder(TEXT("/Game/Characters/SimulatedPlayers/BenchmarkNPC_BP"));
```
improbable-danny, line 33: Should this be static?

improbable-danny, line 33:
```suggestion
	static ConstructorHelpers::FClassFinder<APawn> NPCBPClassFinder(TEXT("/Game/Characters/SimulatedPlayers/BenchmarkNPC_BP"));
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -310,17 +368,48 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
+	{
+		return;
+	}
+
+	const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
+	check(Constants);
+
+	// This test respects the initial delay timer in both native and GDK
+	if (!bHasActorCountFailed &&
+		Constants->ActorCheckDelay.IsReady())
```
improbable-danny, line 392: nit: Newline required here?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -32,6 +58,11 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 	UPROPERTY(ReplicatedUsing = OnRepTotalNPCs, BlueprintReadWrite)
 	int32 TotalNPCs;
 
+	virtual void BuildExpectedActorCounts() {};
```
improbable-danny, line 61:
```suggestion
	virtual void BuildExpectedActorCounts() {}
```

Game/Source/GDKTestGyms/Private/CounterComponent.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+	PrimaryComponentTick.bCanEverTick = true;
+}
+
+void UCounterComponent::TickComponent(float DeltaTime, ELevelTick TickType, FActorComponentTickFunction* ThisTickFunction)
+{
+	Super::TickComponent(DeltaTime, TickType, ThisTickFunction);
+
+	Timer -= DeltaTime;
+	if (Timer <= 0.0f)
```
improbable-danny, line 20: This could utilize the latched timer potentially

Game/Source/GDKTestGyms/OffloadingBenchmarkGymGameMode.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "OffloadingBenchmarkGymGameMode.h"
+
+#include "UObject/ConstructorHelpers.h"
+
+DEFINE_LOG_CATEGORY(LogOffloadingBenchmarkGymGameMode);
+
+AOffloadingBenchmarkGymGameMode::AOffloadingBenchmarkGymGameMode()
+{
+	ConstructorHelpers::FClassFinder<AActor> NPCBPClassFinder(TEXT("/Game/OffloadingBenchmark/OffloadingBenchmark_NPC_BP"));
```
improbable-danny, line 11:
```suggestion
	static ConstructorHelpers::FClassFinder<AActor> NPCBPClassFinder(TEXT("/Game/OffloadingBenchmark/OffloadingBenchmark_NPC_BP"));
```

improbable-danny: I think this is the suggested way for Unreal is that right @m-samiec?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: No I just like it is it separates the conditions to make it readable. Style thing I guess.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -164,6 +171,19 @@ void ABenchmarkGymGameMode::ParsePassedValues()
...
+void ABenchmarkGymGameMode::BuildExpectedActorCounts()
+{
+	Super::BuildExpectedActorCounts();
+
+	AddExpectedActorCount(NPCBPClass, TotalNPCs, 1);
+
+	AddExpectedActorCount(SimulatedPlayerBPClass, ExpectedPlayers, 1);
+
+	const int32 TotalDropCubes = TotalNPCs + ExpectedPlayers;
+	const int32 DropCubeCountVariance = TotalDropCubes * 0.15f;
```
m-samiec, line 183: 15% just a finger in air number here?

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
```diff
@@ -7,26 +7,46 @@
...
+	return !bLocked;
+}
+
+void FMetricTimer::SetLock(bool bState)
+{
+	bLocked = bState;
 }
 
-bool FMetricDelay::IsReady() const
+bool FMetricTimer::HasTimerGoneOff() const
```
m-samiec, line 30: Could we set a max time for these? That way we wouldn't need the additional life time timer, and it could be contained in this obj.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -176,13 +220,13 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
 	check(Constants);
 
 	// This test respects the initial delay timer in both native and GDK
-	if (Constants->PlayerCheckMetricDelay.IsReady())
+	if (Constants->PlayerCheckMetricDelay.HasTimerGoneOff())
 	{
 		bHasDonePlayerCheck = true;
```
improbable-danny, line 225: This is no longer used

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: Yeah, we could experiment with going a bit tighter. For now just whatever we deem reasonable. Any ideas?

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
MatthewSandfordImprobable: So HasTimerGoneOff only returns true when we are after an initial delay but before another time. Hmm potentially. That would work. These would no longer be timers. I feel as though this would be harder to understand. I think everyone understands a timer and knows what to expect. If you feel the "code benefit" would outweigh the "understandability" benefit then I will implement this suggestion.

MatthewSandfordImprobable:
Updated the lifetime timers in the yaml files.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
m-samiec: Work checking what the variance is in one of the upper level stress tests - this can be done by just looking at the num of entities in a deployment, as that's the only actor that's continually created/destroyed.

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
m-samiec: It seems like it could be a common pattern - ie. report the metric within this window.
However if this is the only case currently, then I'm fine with it.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -157,6 +193,13 @@ void ABenchmarkGymGameModeBase::Tick(float DeltaSeconds)
 	TickClientFPSCheck(DeltaSeconds);
 	TickPlayersConnectedCheck(DeltaSeconds);
 	TickUXMetricCheck(DeltaSeconds);
+	TickActorCountCheck(DeltaSeconds);
+
+	if (HasAuthority() &&
+		PrintMetricsTimer.HasTimerGoneOff())
```
m-samiec, line 199: This logic seems a little confusing - we're checking if it's "gone off" - but then just immediately resetting it. Why isn't the print logic tied up here?

m-samiec, line 199: This logic seems a little confusing - we're checking if it's "gone off" - but then just immediately resetting it. Why isn't the print logic tied up here?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -176,13 +219,13 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
...
 	// This test respects the initial delay timer in both native and GDK
-	if (Constants->PlayerCheckMetricDelay.IsReady())
+	if (Constants->PlayerCheckMetricDelay.HasTimerGoneOff())
 	{
 		bHasDonePlayerCheck = true;
 		if (ActivePlayers < RequiredPlayers)
 		{
 			// This log is used by the NFR pipeline to indicate if a client failed to connect
-			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("A client connection was dropped. Required %d, got %d, num players %d"), RequiredPlayers, ActivePlayers, GetNumPlayers());
+			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("NFR scenario failed: Client connection dropped. Required %d, got %d, num players %d"), RequiredPlayers, ActivePlayers, GetNumPlayers());
```
m-samiec, line 228: nit: could we make "NFR scenario failed" a const string that can get used everywhere?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -414,4 +510,25 @@ void ABenchmarkGymGameModeBase::SetTotalNPCs(int32 Value)
...
+{
+	const UCounterComponent* CounterComponent = Cast<UCounterComponent>(GameState->GetComponentByClass(UCounterComponent::StaticClass()));
+	check(CounterComponent != nullptr)
+
+	return CounterComponent->GetActorClassCount(ActorClass);
+}
+
+void ABenchmarkGymGameModeBase::SetLifetime(int32 Lifetime)
+{
+	if (TestLifetimeTimer.SetTimer(Lifetime))
```
m-samiec, line 525: Was this getting called multiple times? What's the value to not allowing it to be reset?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -307,20 +350,55 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
 	}
 }
 
+void ABenchmarkGymGameModeBase::TickActorCountCheck(float DeltaSeconds)
+{
+	const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
+	check(Constants);
+
+	// This test respects the initial delay timer in both native and GDK
+	if (!bHasActorCountFailed &&
```
m-samiec, line 371: I thought we agreed to do this all the time (ie. set bHasActorCountFailed) but only actually fail the scenario when the timer has expired?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -414,4 +510,25 @@ void ABenchmarkGymGameModeBase::SetTotalNPCs(int32 Value)
...
+void ABenchmarkGymGameModeBase::SetLifetime(int32 Lifetime)
+{
+	if (TestLifetimeTimer.SetTimer(Lifetime))
+	{
+		TestLifetimeTimer.SetLock(true);
+		UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("Setting NFR test lifetime %d"), Lifetime);
+	}
+	else
+	{
+		UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("Could not set NFR test liftime to %d. Timer was locked."), Lifetime);
```
m-samiec, line 532: This seems like a warning currently

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: So the timer is just being reset so it is similar to a looping timer. Each "tick expectation" function has the chance to check if the timer went off and then print stats if it wants to. We could tie the print logic at this point but we would have to somehow get the relevant stats to print. These would all have to be member variables (some are already). That just didn't seem neat to me. Members for the purpose of passing information around just for the sake of logs.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Sure! I like it!

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Yes it get's called multiple times but actually... maybe the solution is to not call it twice. I have a feeling it "needs" to be called twice but I will check that.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: We still get the print out of how many actors are in the scenario as that is done by the counter component on the game state. The only purpose of further logic is to print the failure. Unless I am missing something?

Game/Source/GDKTestGyms/BenchmarkGymNPCSpawner.cpp
```diff
@@ -9,16 +9,11 @@ DEFINE_LOG_CATEGORY(LogBenchmarkGymNPCSpawner);
...
-	if (NPCBPClass.Class != NULL)
-	{
-		NPCPawnClass = NPCBPClass.Class;
-	}
 	SetReplicates(true);
 	bAlwaysRelevant = true;
 }
 
-void ABenchmarkGymNPCSpawner::CrossServerSpawn_Implementation(const FVector& SpawnLocation, const FBlackboardValues& BlackboardValues)
+void ABenchmarkGymNPCSpawner::CrossServerSpawn_Implementation(TSubclassOf<APawn> NPCClass, const FVector& SpawnLocation, const FBlackboardValues& BlackboardValues)
```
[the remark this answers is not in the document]
MatthewSandfordImprobable: Originally the spawner just spawned a class the was defined in the spawners constructor. This means we had two bits in the code setting which class we wanted to use for the NPCClass. Now the spawner is told what called to spawn so it's only defined in one place. I could add an RPC to just set the class for the spawner to spawn if you feel that would be an improvement. It is somewhat redundant to keep sending the same class via RPC.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Duplicate

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: Looking at graphana metrics I saw a roughly 4.5% variance in two deployments. I have set the allowed variance to 8% to be safe.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: It's also to register the metric back to grafana - which won't be done - or it will be done but with a potentially "fake" value.

Game/Source/GDKTestGyms/BenchmarkGymNPCSpawner.cpp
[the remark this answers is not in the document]
m-samiec: No this is fine

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: Okay - as there's a dependency on call order here could you add a brief comment detailing this usage - ie. give each tick function that ability to print using this timer, then reset.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Ahh yes! Will change.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Gonna keep this code for now. Ran a tests trying to fix this but it failed and I don't really want to spend time looking into it right now. I think this code is acceptable for now but I will create a follow up ticket to work out why we call the parseparameters a second time. Let me know if that is good with you.

PR #140, opened 2020-07-07
Added the map that runs the character movement test

AndreiTeodorLazar:
This PR closes the [EngineNetTestPR](https://github.com/improbable/UnrealGDKEngineNetTest/pull/69)  and adds the character movement test in the new framework


ImprobableNic:
So is the plan to get rid of the EngineNetTest project, and only rely on the Gyms ?

mironec:
Yes, mostly - consolidates testing into one project, and long-term we should have more automated tests here than manual ones.
The only real pain here is compatibility with multiple engine versions, but solved hackily for now.

PR #158, opened 2020-07-27
CrossServerRPC Test Map

AndreiTeodorLazar:
This PR adds the map that contains the CrossServerRPC Test and updates the user manual to reflect that an automated version of the gym is now available.


oblm:
I recommend changing the note to:

##### Server to server RPC gym
* NOTE: This gym can be run both as an automated test and a manual one.
* Automation steps
In order to run the test, follow the steps:
  1. Open the Session Frontend: Window -> Developer Tools -> Session Frontend.
  2. On the Automation tab, search for SpatialTestCrossServerRPC1, tick the box corresponding to it and hit Start Tests.
  3. The Session Frontend will then prompt you with the result of the test.
* Manual Steps:
ToDO: @AndreiTeodorLazar add manual steps that end with:
* If it is working correctly, the normally white cubes will start flashing the colours of the other workers in the level, and the counters above the cubes will turn the corresponding worker colours and increment for each RPC received. If not, the cubes will timeout waiting for RPCs to be received and this will be indicated above the cubes.

* This gym demonstrates that:
  * Actors owned by different servers correctly send server-to-server RPCs.
* This gym contains:
  * A set of cubes placed in four quadrants of the level which:
    * Randomly send RPCs from each worker to the other cubes in the level.
    * Change colour to indicate which worker the owning worker just received an RPC from (where colours match the inspector clours used by the spatial debugger).
    * Show a count of how many RPCs have been received which is shown above the cube.

PR #167, opened 2020-07-31
Improving drop cube variance.

MatthewSandfordImprobable:


Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -174,7 +174,7 @@ void ABenchmarkGymGameMode::BuildExpectedActorCounts()
 	Super::BuildExpectedActorCounts();
 
 	const int32 TotalDropCubes = TotalNPCs + ExpectedPlayers;
-	const int32 DropCubeCountVariance = FMath::CeilToInt(TotalDropCubes * 0.1f);
```
improbable-danny, line 177: Is this still failing at low numbers? I wonder if 10% isn't a good model of how much variance there is

improbable-danny: Maybe a higher constant would be better (like 5)? The significance of the constant tails off at higher numbers and gives more lenience in lower tests

MatthewSandfordImprobable: ^ Exactly. It's a balance between the constant and the percentage. I think 5 is a bit high from what I have seen. 2 might be more appropriate but for now I wanted to be conservative and just add 1 as that would have fixed the only failure I have seen.

improbable-danny: Sounds good, maybe we should use 2? I'm more worried about the cases we don't see/inspect but I guess it'll be pretty obvious

MatthewSandfordImprobable: Changed to 2.

PR #170, opened 2020-08-06
Modified the ReplicatedStartupActor Map.

AndreiTeodorLazar:
This PR modifies the ReplicatedStartupActor map to contain the new class ReplicatedStartupActor, instead of the ReplicatedTestActorBase, in order to allow for testing that properties are replicated correctly. 

Before the change, the map contained:
image

After the change, the differences in the map are: The plane was significantly scaled, and the startup actor is now of a different class, as seen in the picture below:
image


mironec:
conflict on the map, will have to be resolved

PR #172, opened 2020-08-12
Fixing actor delegation test

nafonso:
There was a reversed branch logic (ie false output should be true and vice versa)


mironec:
screenshot maybe?

nafonso:
> screenshot maybe?

This is how it was before:
image

After the change it's this:
image

The bug was related to remove working via RPC in the past which would push the removal by test auth instead of actor auth (a bug that stayed from the refactor that happened on that system)

PR #174, opened 2020-08-13
Feature/unr 4028 has authority tests

nafonso:
Test Map for SpatialAuthorityTest.  Also fixes 2 zoning maps not have proper multi worker settings.


mironec:
Ah, sorry - could you also move the SpatialNetworkingMap and SpatialZoningMap into the CI_Fast folder so they get run on every commit?

nafonso:
> Ah, sorry - could you also move the SpatialNetworkingMap and SpatialZoningMap into the CI_Fast folder so they get run on every commit?

Done

PR #177, opened 2020-08-19
Added the Map for the RepNotify test

AndreiTeodorLazar:
This PR adds a Map for the RepNotify test.


AndreiTeodorLazar:
I think this set should not be prone to flakes and it also runs super quick (about 0.7 seconds on my machine), do you think this should be moved to the SpatialNetworkingMap and added as part of the pre-merge CI?

mironec:
> I think this test should not be prone to flakes and it also runs super quick (about 0.7 seconds on my machine), do you think this should be moved to the SpatialNetworkingMap and added as part of the pre-merge CI?
Ah yes, that was the original intention: but beware, you have to modify the SpatialNetworking map on 4.24 to pass CI. Hm... I should add CI to the testgyms...

AndreiTeodorLazar:
Just a note that I have moved the RepNotify test to the Networking Map.

PR #183, opened 2020-08-25
Test/spatial test rep notify

AndreiTeodorLazar:
This PR is used add the RepNotify test to the NetworkingMap and to match the name of the corresponding GDK branch.


mironec (approved):
As long as the GDK CI passes

PR #186, opened 2020-09-03
UNR-4043 Ensure in zoned test cases player density

jayprobable:
This PR changes how spawn points are distributed to ensure that 25% of the spawn points reside near/on a zone boundary. For a single worker test case the behavior will remain as before with a grid of spawn points the center of which resides at 0,0.

The number of workers is defined via a worker flag: num_workers, the default value is 1.

Because the GridBasedLBStrategy actually switches X and Y, this code also swaps X and Y when laying out the spawn points.
See the Init() function of the GridBasedLBStrategy for more context.

The key concept here is that we have a predetermined world size of 10km^2, based on this the zones are divided and the boundaries are computed, we use 10km because it's large enough that we _should_ always have large enough zones for the number of players we seek to test.

Fundamentally, we need to know configuration for the load-balancing before we necessarily have all the worker flags, if we were to manage these parameters via commandline options this could be much simpler and easier to control variables, such as the world size/zone sizes. Currently because the launch arguments are embedded into a descriptor protobuf and loaded as an assembly asset modifying arguments at launch time (e.g. leveraging SSE dimensions) is not possible. This is an artifact of the legacy platform design. In the new (IMP) platform design hopefully we will account for this inflexibility (I will reach out to appropriate parties to explain the reasons why the ability to dynamically add launch arguments without needing to create a separate assembly is highly desirable).

This has been tested by running cloud deployments manually.
As well here is the nfr test:
https://buildkite.com/improbable/product-research-benchmarks-run/builds/15129

Associated TestGymBuildKite PR: https://github.com/improbable/TestGymBuildKite/pull/175

Screen Shot 2020-09-03 at 4 29 02 PM


Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -211,19 +225,65 @@ void ABenchmarkGymGameMode::GenerateGridSettings(int DistBetweenPoints, int NumP
 void ABenchmarkGymGameMode::GenerateSpawnPointClusters(int NumClusters)
 {
 	const int DistBetweenClusterCenters = 40000; // 400 meters, in Unreal units.
-	int NumRows, NumCols, MinRelativeX, MinRelativeY;
-	GenerateGridSettings(DistBetweenClusterCenters, NumClusters, NumRows, NumCols, MinRelativeX, MinRelativeY);
 
-	UE_LOG(LogBenchmarkGymGameMode, Log, TEXT("Creating player cluster grid of %d rows by %d columns"), NumRows, NumCols);
-	for (int i = 0; i < NumClusters; i++)
+	//we use a fixed size 10km
```
MatthewSandfordImprobable, line 229: Nit: Usually start comment with a space and capital letter.

MatthewSandfordImprobable (approved):

yujunting (approved):

PR #191, opened 2020-09-11
UNR-4226 Change UnrealTotalAuthoritativePlayers to UnrealActivePlayers

yujunting:
To keep the nfr testing with the same expectations both on zoning and none-zoning.
Add a new metric is not necessary.
------------------------------------------
Count the authoritative players from UUserExperienceReporter with HasAuthority() check.
GetNumPlayers() is necessary for analyzing the log of UnrealWorker, So I restore it back.

Jira ticket:
https://improbableio.atlassian.net/browse/UNR-4226

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -197,7 +197,6 @@ void ABenchmarkGymGameModeBase::Tick(float DeltaSeconds)
 	TickPlayersConnectedCheck(DeltaSeconds);
 	TickUXMetricCheck(DeltaSeconds);
 	TickActorCountCheck(DeltaSeconds);
-	ReportAuthoritativePlayers(FPlatformProcess::ComputerName(), GetAuthoritativePlayers());
 	
 	// PrintMetricsTimer needs to be reset at the the end of ABenchmarkGymGameModeBase::Tick.
 	// This is so that the above function have a chance to run logic dependant on PrintMetricsTimer.HasTimerGoneOff().
@@ -305,11 +304,6 @@ void ABenchmarkGymGameModeBase::TickClientFPSCheck(float DeltaSeconds)
 
 void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
 {
-	if (!HasAuthority())
-	{
-		return;
-	}
-
 	int UXComponentCount = 0;
 	int ValidRTTCount = 0;
 	int ValidUpdateTimeDeltaCount = 0;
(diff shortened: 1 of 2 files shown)


m-samiec:
> removed GetNumPlayers() from one of the logs is I think it's not precise enough for client connections.

Could you clarify what you mean here?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -112,8 +111,7 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 
 	double GetClientRTT() const { return AveragedClientRTTSeconds; }
 	double GetClientUpdateTimeDelta() const { return AveragedClientUpdateTimeDeltaSeconds; }
-	double GetPlayersConnected() const { return ActivePlayers; }
-	double GetTotalAuthoritativePlayers() const;
+	double GetPlayersConnected() const { return (double)ActivePlayers; }
```
jayprobable, line 114: I prefer explicit c++ casts were used instead of c-style casting ... e.g. static_cast<double>(ActivePlayers) .. though is it even necessary? implicit conversion from int32 to double should happen here.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -238,12 +231,12 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
 		if (ActivePlayers < RequiredPlayers)
 		{
 			// This log is used by the NFR pipeline to indicate if a client failed to connect
-			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Client connection dropped. Required %d, got %d, num players %d"), *NFRFailureString, RequiredPlayers, ActivePlayers, GetNumPlayers());
+			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Client connection dropped. Required %d, got %d"), *NFRFailureString, RequiredPlayers, ActivePlayers);
 		}
 		else
 		{
 			// Useful for NFR log inspection
-			NFR_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("All clients successfully connected. Required %d, got %d, num players %d"), RequiredPlayers, ActivePlayers, GetNumPlayers());
```
jayprobable, line 246: There's no reason to remove num players from this log line in this PR.

jayprobable (changes requested):
Should we be counting the authoritative UXComponent count  in order to mantain parity with what the code was doing previously? What's the difference in this case? Also to @m-samiec point, what is wrong with GetNumPlayers()?

jayprobable (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -111,7 +111,7 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 
 	double GetClientRTT() const { return AveragedClientRTTSeconds; }
 	double GetClientUpdateTimeDelta() const { return AveragedClientUpdateTimeDeltaSeconds; }
-	double GetPlayersConnected() const { return (double)ActivePlayers; }
+	double GetPlayersConnected() const { return ActivePlayers; }
```
MatthewSandfordImprobable, line 113: Think we should static cast ActivePlayers to double

MatthewSandfordImprobable (approved):

PR #192, opened 2020-09-11
Event tracing functional test gym

MatthewSandfordImprobable:
The added map is meant to be very generic. It simply has a character that moves around periodically. 
Relies on: https://github.com/spatialos/UnrealGDK/pull/2531

Changes:
Added basic map that contains all the event tracing functional tests.
Added a game mode for the map to use that simply specifies a character.
Added a simple character that moves around periodically blueprint:
image


mironec:
I would normally advocate for moving tests fully to C++, but I don't think this particular test will be a part of CI, so may be okay.

mironec (approved):

improbable-danny (approved):
Looks good!

PR #193, opened 2020-09-17
UNR-4045 Add UnrealActorMigration as a new metric for zoning-nfr 

yujunting:
For zoning specific overhead, add a new metric UnrealActorMigration.
1) Count all the Actors which HasAuthority() in each tick
2) Calculate the delta with the count of the last frame
3) sum all the abs of delta in 5 minutes as MigrationOfCurrentWorker
4) sum all the MigrationOfCurrentWorker from all workers as the value of the metric

Added a new WorkerFlag that changes how Migration is verified and prints the associated logs in case of failure.
Changed Migration to every frame instead of the previous every 9000 frames.
Added DelayTimer to avoid statistical errors caused by the players not being fully connected.
Changed the reporting of Migration to only for changes.

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -75,7 +75,7 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, MigrationCountSeconds(0.0)
 	, MigrationWindowSeconds(5*60.0f)
 	, MinActorMigrationPerSecond(0.0)
-	, ActorMigrationCheckTimer(6*60) // 1-minute later then UNFRConstants::ActorMigrationCheckDelay to make sure all the workers had reported their migration
+	, ActorMigrationCheckTimer(11*60) // 1-minute later then UNFRConstants::ActorMigrationCheckDelay + MigrationWindowSeconds to make sure all the workers had reported their migration
 	, ActivePlayerReportDelayTimer(5*60)
 	, PrintMetricsTimer(10)
 	, TestLifetimeTimer(0)
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -606,9 +606,8 @@ void ABenchmarkGymGameModeBase::TickActorMigration(float DeltaSeconds)
 		if (MigrationCountSeconds > MigrationWindowSeconds)
 		{
 			MigrationDeltaPair OldestValue;
-			if (MigrationDeltaHistory.Peek(OldestValue))
+			if (MigrationDeltaHistory.Dequeue(OldestValue))
 			{
-				MigrationDeltaHistory.Pop();
 				MigrationOfCurrentWorker -= OldestValue.Key;
 				MigrationSeconds -= OldestValue.Value;
 				bChanged = (Delta != OldestValue.Key);
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -5,7 +5,6 @@
 #include "CounterComponent.h"
 #include "Engine/World.h"
 #include "EngineClasses/SpatialNetDriver.h"
-#include "GameFramework/Character.h"
 #include "GameFramework/GameStateBase.h"
 #include "GameFramework/MovementComponent.h"
 #include "GDKTestGymsGameInstance.h"
@@ -38,7 +37,7 @@ namespace
 	const FString TestLiftimeWorkerFlag = TEXT("test_lifetime");
 	const FString TestLiftimeCommandLineKey = TEXT("-TestLifetime=");
 
-	const FString HandOverTotalFrameCountWorkerFlag = TEXT("HandOver_total_frame_count");
+	const FString HandOverTotalFrameCountWorkerFlag = TEXT("handover_total_frame_count");
 	const FString TotalPlayerWorkerFlag = TEXT("total_players");
 	const FString TotalNPCsWorkerFlag = TEXT("total_npcs");
 	const FString RequiredPlayersWorkerFlag = TEXT("required_players");
@@ -70,7 +69,7 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, LastTimeHandOverActors(0)
 	, HandOverdActorsOfCurrentWorker(0)
 	, HandOverFrameCount(0)
-	, HandOverTotalFrameCount(9000)
+	, HandOverTotalFrameCount(9000) // About 5Mins(FPS=30 on worker)
 	, PrintMetricsTimer(10)
 	, TestLifetimeTimer(0)
 {
@@ -621,6 +620,5 @@ double ABenchmarkGymGameModeBase::GetTotalHandOverActors() const
 	{
 		TotalHandOverActors += kv.Value;
 	}
-	UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("TotalHandOverActors=%d"), TotalHandOverActors);
 	return static_cast<double>(TotalHandOverActors);
 }
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -626,6 +626,7 @@ void ABenchmarkGymGameModeBase::TickActorMigration(float DeltaSeconds)
 			}
 		}
 		MigrationOfCurrentWorker += Delta;
+		MigrationExactlyWindowSeconds += DeltaSeconds;
 		PreviousTickMigration = UXAuthActorCount;
 
 		if (MigrationCountSeconds > MigrationWindowSeconds)
@@ -649,15 +650,15 @@ void ABenchmarkGymGameModeBase::TickActorMigration(float DeltaSeconds)
 					if (AverageActorMigration < MinActorMigrationPerSecond)
 					{
 						bHasActorMigrationCheckFailed = true;
-						NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Actor migration check failed. TotalMigrations=%.3f AverageActorMigration=%.3f MinActorMigration=%.3f"),
-							*NFRFailureString, TotalMigrations, AverageActorMigration, MinActorMigrationPerSecond);
+						NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Actor migration check failed. TotalMigrations=%.3f AverageActorMigration=%.3f MinActorMigration=%.3f MigrationExactlyWindowSeconds=%.3f"),
+							*NFRFailureString, TotalMigrations, AverageActorMigration, MinActorMigrationPerSecond, MigrationExactlyWindowSeconds);
 					}
 					else
 					{
 						// reset timer for next check after 10s
 						ActorMigrationCheckTimer.SetTimer(10);
-						UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("Actor migration check TotalMigrations=%.3f AverageActorMigration=%.3f MinActorMigration=%.3f"),
-							TotalMigrations, AverageActorMigration, MinActorMigrationPerSecond);
+						UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("Actor migration check TotalMigrations=%.3f AverageActorMigration=%.3f MinActorMigration=%.3f MigrationExactlyWindowSeconds=%.3f"),
+							TotalMigrations, AverageActorMigration, MinActorMigrationPerSecond, MigrationExactlyWindowSeconds);
 					}
 				}
 			}
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -72,9 +72,9 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, PreviousTickMigration(0)
 	, UXAuthActorCount(0)
 	, MigrationOfCurrentWorker(0)
-	, MigrationFrameCount(0)
-	, MigrationWindowFrameCount(9000) // About 5Mins(FPS=30 on worker)
-	, MinActorMigration(0.0)
+	, MingrationCountSeconds(0.0)
+	, MigrationWindowSeconds(5*60.0f)
+	, MinActorMigrationPerSecond(0.0)
 	, ActorMigrationCheckTimer(10)
 	, PrintMetricsTimer(10)
 	, TestLifetimeTimer(0)
@@ -449,7 +449,7 @@ void ABenchmarkGymGameModeBase::ParsePassedValues()
 
 		FParse::Value(*CommandLine, *MaxRoundTripCommandLineKey, MaxClientRoundTripSeconds);
 		FParse::Value(*CommandLine, *MaxUpdateTimeDeltaCommandLineKey, MaxClientUpdateTimeDeltaSeconds);
-		FParse::Value(*CommandLine, *MinActorMigrationCommandLineKey, MinActorMigration);
+		FParse::Value(*CommandLine, *MinActorMigrationCommandLineKey, MinActorMigrationPerSecond);
 	}
 	else if (GetDefault<UGeneralProjectSettings>()->UsesSpatialNetworking())
 	{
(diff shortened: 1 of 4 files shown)


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -34,6 +37,7 @@ namespace
 	const FString TestLiftimeWorkerFlag = TEXT("test_lifetime");
 	const FString TestLiftimeCommandLineKey = TEXT("-TestLifetime=");
 
+	const FString HandOverTotalFrameCountWorkerFlag = TEXT("handover_total_frame_count");
```
MatthewSandfordImprobable, line 40: Do we use this?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -6,9 +6,11 @@
 #include "Engine/World.h"
 #include "EngineClasses/SpatialNetDriver.h"
 #include "GameFramework/GameStateBase.h"
+#include "GameFramework/MovementComponent.h"
 #include "GDKTestGymsGameInstance.h"
 #include "GeneralProjectSettings.h"
 #include "Interop/SpatialWorkerFlags.h"
+#include "Kismet/GameplayStatics.h"
```
MatthewSandfordImprobable, line 13: Is this used?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -70,6 +70,9 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 	virtual void ReportAuthoritativePlayers(const FString& WorkerID, int AuthoritativePlayers);
 	void ReportAuthoritativePlayers_Implementation(const FString& WorkerID, int AuthoritativePlayers);
 
+	UFUNCTION(CrossServer, Reliable)
+	virtual void ReportHandOverActors(const FString& WorkerID, int HandOverActors);
```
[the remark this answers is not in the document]
yujunting: 😄

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -87,6 +88,14 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 	bool bExpectedActorCountsInitialised;
 	int32 ActivePlayers; // All authoritative players from all workers
 
+	// For actor migration count
+	int32 PreviousTickMigration;
+	TArray<int32> ToBeRemovedMigration;
```
jayprobable, line 93: Prefer naming this something like MigrationDeltas, also we should be dividing the Migration count by the number of frames in the array so that we are getting an average per tick of migrating players. From this metric we can represent a minimum percentage of the total number of players as an expectation.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -555,3 +568,55 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayers_Implementation(const
 		}
 	}
 }
+
+void ABenchmarkGymGameModeBase::TickActorMigration(float DeltaSeconds)
+{
+	// Count how many actors hand over authority in 1 tick
+	int MovementAuthActors = GetMovementAuthActors();
+	int Delta = FMath::Abs(MovementAuthActors - PreviousTickMigration);
```
jayprobable, line 576: The first time this is called the PreviousTickMigration will be zero... in fact, any time players connect they will cause the MovementAuthActors to increment.. as long as we ensure we don't check this metric too early I think it's fine.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -105,12 +114,15 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
...
+	void TickActorMigration(float DeltaSeconds);
 
+	int GetMovementAuthActors() const;
 	void SetTotalNPCs(int32 Value);
 
 	double GetClientRTT() const { return AveragedClientRTTSeconds; }
 	double GetClientUpdateTimeDelta() const { return AveragedClientUpdateTimeDeltaSeconds; }
 	double GetPlayersConnected() const { return static_cast<double>(ActivePlayers); }
+	double GetTotalMigrations() const;
 	double GetFPSValid() const { return !bHasFpsFailed ? 1.0 : 0.0; }
```
jayprobable, line 126: Are we going to add an GetMigrationsValid() method?

jayprobable (commented):
I think that the number that we report should be the average number of players migrating per tick (which can be averaged over the 9000 frames) then from this we can define the lower limit before a test scenario is invalid based on a percentage of the number of players.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
yujunting: Yes, only keep a GetTotalMigrationValid() is better.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -62,6 +66,10 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, bActorCountFailureState(false)
 	, bExpectedActorCountsInitialised(false)
 	, ActivePlayers(0)
+	, LastTimeHandOverActors(0)
+	, HandOverdActorsOfCurrentWorker(0)
```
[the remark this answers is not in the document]
yujunting: I had already changed the variable name as migration

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -62,6 +66,10 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
...
+	, HandOverFrameCount(0)
+	, HandOverTotalFrameCount(9000) // About 5Mins(FPS=30 on worker)
```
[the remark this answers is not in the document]
yujunting: I had already changed the variable name as migration

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -62,6 +66,10 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, bActorCountFailureState(false)
 	, bExpectedActorCountsInitialised(false)
 	, ActivePlayers(0)
+	, LastTimeHandOverActors(0)
```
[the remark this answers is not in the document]
yujunting: yes, previous is better.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -70,6 +70,9 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 	virtual void ReportAuthoritativePlayers(const FString& WorkerID, int AuthoritativePlayers);
 	void ReportAuthoritativePlayers_Implementation(const FString& WorkerID, int AuthoritativePlayers);
 
+	UFUNCTION(CrossServer, Reliable)
+	virtual void ReportHandOverActors(const FString& WorkerID, int HandOverActors);
+	void ReportHandOverActors_Implementation(const FString& WorkerID, int HandOverActors);
```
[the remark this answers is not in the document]
yujunting: yes, void ReportHandOverActors_Implementation is not necessary.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -87,6 +90,14 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 	bool bExpectedActorCountsInitialised;
 	int32 ActivePlayers; // All authoritative players from all workers
 
+	// For hand over authoritative actors count
+	int LastTimeHandOverActors;
```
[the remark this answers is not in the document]
yujunting: thanks, change it to int32 is better

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
yujunting: I use `ActorMigrationCheckDelay` for it

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -425,11 +445,14 @@ void ABenchmarkGymGameModeBase::ParsePassedValues()
 
 		FParse::Value(*CommandLine, *MaxRoundTripCommandLineKey, MaxClientRoundTripSeconds);
 		FParse::Value(*CommandLine, *MaxUpdateTimeDeltaCommandLineKey, MaxClientUpdateTimeDeltaSeconds);
+		float fValue = 0.0f;
```
jayprobable, line 448: fValue, doesn't follow Unreal naming convention... Value of What? pseudo hungarian notation is unnecessary

jayprobable: Why do we need this temporary at all, just pass MinActorMigration

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -555,3 +586,91 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayers_Implementation(const
...
+						// reset timer for next check after 10s
+						ActorMigrationCheckTimer.SetTimer(10);
+					}
+				}
+			}
+		}
+	}
+}
+
+int ABenchmarkGymGameModeBase::GetMovementAuthActors() const
```
jayprobable, line 655: The name of this function is misleading, What does this function return? A count of actor which have a UUserExperienceComponent? .. Perhaps GetAuthUxActorsCount() or GetCountAuthUxActors() ...? Is this function new or do we already have a function that gets the count of authoritative actors with UxComponent..?

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
```diff
@@ -47,6 +47,7 @@ UNFRConstants::UNFRConstants()
 	, ClientFPSMetricDelay(60)
 	, UXMetricDelay(10 * 60)
 	, ActorCheckDelay(10 * 60)
+	, ActorMigrationCheckDelay(5 * 60)
```
jayprobable, line 50: Why is the delay not consistent with the UXMetricDelay and the ActorCheckDelay?

jayprobable (changes requested):

Game/Source/GDKTestGyms/Private/NFRConstants.cpp
yujunting: If we have 10 minutes delay check for migration, and the migration need 5 minutes(9000 ticks) for statistic, I'm worried about it didn't have enough time to report the value before end.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
yujunting: no, MinActorMigration is double, but `FParse::Value` doesn't support double. So we need to use the temporary variable to convert it.

jayprobable (approved):

MatthewSandfordImprobable (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -62,6 +68,14 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, bActorCountFailureState(false)
 	, bExpectedActorCountsInitialised(false)
 	, ActivePlayers(0)
+	, bHasActorMigrationCheckFailed(false)
+	, PreviousTickMigration(0)
+	, UXAuthActorCount(0)
+	, MigrationOfCurrentWorker(0)
+	, MigrationFrameCount(0)
+	, MigrationWindowFrameCount(9000) // About 5Mins(FPS=30 on worker)
```
m-samiec, line 76: Why are we using ticks and not seconds/milli-seconds?
Character movement is time based, and when server frame rate struggles (we allow framerate to drop to 20 fps before failing), the number of ticks will be fewer.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -87,6 +88,18 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
...
+	// For actor migration count
+	bool bHasActorMigrationCheckFailed;
+	int32 PreviousTickMigration;
+	TArray<int32> ToBeRemovedMigrationDeltas;
+	int32 UXAuthActorCount;
+	int32 MigrationOfCurrentWorker;
+	int32 MigrationFrameCount;
+	int32 MigrationWindowFrameCount;
+	TMap<FString, int32>	MapWorkerActorMigration;
+	float MinActorMigration;
```
m-samiec, line 100: nit - include unit in these types of variables

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
yujunting: If an Actor moves like this over a short period of time, A->B->A, it's easy to miss statistics if use Seconds or Milliseconds.
The comment 5Mins only means about, not an exact time interval.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
yujunting: Would you please explain it clearly for me? or give me an example like how to include a unit for MigrationFrameCount or other variables?

m-samiec: ie. MinActorMigrationPerSecond or MinActorMigrationPerMillisecond

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -555,3 +588,83 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayers_Implementation(const
...
+	if (Constants->ActorMigrationCheckDelay.HasTimerGoneOff())
+	{
+		// Count how many actors hand over authority in 1 tick
+		int Delta = FMath::Abs(UXAuthActorCount - PreviousTickMigration);
+		ToBeRemovedMigrationDeltas.Enqueue(ToBeRemovedDelta(Delta, DeltaSeconds));
+		bool bChanged = false;
+		if (MigrationCountSeconds > MigrationWindowSeconds)
+		{
+			ToBeRemovedDelta RemoveValue;
+			if (ToBeRemovedMigrationDeltas.Peek(RemoveValue))
```
jayprobable, line 612: Prefer use Dequeue(OldestValue) ... instead of peek then pop, also prefer the name OldestValue instead of RemoveValue, OldestValue is more clear.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -87,6 +88,20 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
 	bool bExpectedActorCountsInitialised;
 	int32 ActivePlayers; // All authoritative players from all workers
 
+	// For actor migration count
+	bool bHasActorMigrationCheckFailed;
+	int32 PreviousTickMigration;
+	typedef TTuple<int32, float> ToBeRemovedDelta;
```
jayprobable, line 94: Is this typedef necessary? At the least change the type name, MigrationDeltaPair (for example), Generally speaking naming things ToBeXXXXX feels clumsy, I prefer naming containers based on what they contain, not how they will be used.. e.g. MigrationDeltas ... if it's necessary to decribe the usage a comment should suffice

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -555,3 +588,83 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayers_Implementation(const
...
+		}
+		MigrationCountSeconds += DeltaSeconds;
+	}
+}
+
+void ABenchmarkGymGameModeBase::ReportMigration_Implementation(const FString& WorkerID, const float AverageMigration)
+{
+	if (HasAuthority())
+	{
+		float& Value = MapWorkerActorMigration.FindOrAdd(WorkerID);
```
jayprobable, line 667: I didn't ask this in the PR for connected clients, but why use FindOrAdd?
Why not use .Emplace? operator [] ..

jayprobable (changes requested):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
yujunting: There are 2 places to use the type TTuple<int32, float>, so I make a typedef.
I will change the name to MigrationDeltaPair

jayprobable (approved):

PR #198, opened 2020-09-23
fix long debug record compilation issues

jacquese:
had a bunch of problems with building the GDK test gyms because of this legacy property thing. This PR cleans up includes such that we can turn it off, and then also turns it off. I no longer get "debug compilation record too long" errors when compiling now. (at least, so far)


m-samiec (approved):
Oh, nice catch

PR #204, opened 2020-09-24
UNR-4163 zoning latency nfr

jayprobable:
This PR contains blueprint changes to create CrossServer RPC to CrossServer RPC and CrossServer RPC to Property Update latency measurements.

image

Above is part of the core change. OnAuthorityGained we enable timers to send CrossServer latency traces.
A few key points:

RPC to RPC
1. We find a target on another worker (by enumerating all the latency player characters) and checking for not HasAuthority()..
2. We make a cross server rpc call passing our own latency player character as a target...
3. The player on the other worker then calls a cross server rpc on our player to close the loop.

RPC to Property Update
1. We find a target on another worker (same as above)..
2. We make a cross server rpc call...
3. The player on the other worker then updates a property with replication and a repnotify event
4. The rep notify handler checks for "remote" and then ends the trace... (by the way tracing works, when the originating worker ends the trace, the results will be recorded).

How did I test this, primarily I ran cloud deployments with 15 sim players and a density of 3 players, I used the benchmark gym but overrode the game mode to be the latency game mode. I set the worker flags to have expectations as if the nfr framework had started it (but obvious with my player count of 15 and density of 3, etc).. I dug through logs and added plenty of logging (though for this final PR I've omitted it)..

I also ran a bunch of test cases using the nfr framework, until finally I got to this success case:
https://buildkite.com/improbable/product-research-benchmarks-run/builds/15865#_

This PR requires https://github.com/spatialos/UnrealGDK/pull/2575 be merged first.


improbable-danny:
Any thoughts on the latency being so high for cross-server? Looks like the latency tracer is getting into an invalid state and blocking the subsequent traces. Maybe it's worth mitigating this if its breaking some stuff? 

Also see this in one of the worker logs 
`[2020.09.24-11.47.22:907][576]LogSpatialReceiver: Warning: Executed RPC SimulatedLatencyPlayerCharacter_BP_C_2147482523::OnInput_CrossServerCall with unresolved references ((entity ID: 8199, offset: 0)) after 1.270 seconds of queueing.  Owner name: PersistentLevel`
Might explain the latency!

improbable-danny:
Had a quick look, it might be getting into an invalid state because the non-authoritive worker is ending the trace where we expect the trace starting client to end the trace. This was also partly done to avoid clock sync issues between different server boxes!

MatthewSandfordImprobable:
What is the functionality of "IsCrossServerTraceEnabled"?

MatthewSandfordImprobable:
Otherwise seems good to me

MatthewSandfordImprobable (approved):

improbable-danny:
I think we can frame these latency numbers better after looking at the traces. We used shooting as a model for the Input > Server RPC > Client RPC, something we want to track is probably server-validated shooting an actor on another server (so Input > Server RPC > Auth Server RPC > Client RPC). Our Actor::TakeDamage is cross-server by default so likely something that will be present in zoned games (and the same case for property updates) 

It gives us an easy base to compare with single-worker too which is really handy as it introduces an additional runtime/worker hop

Thoughts? cc @m-samiec @MatthewSandfordImprobable

MatthewSandfordImprobable:
> I think we can frame these latency numbers better after looking at the traces. We used shooting as a model for the Input > Server RPC > Client RPC, something we want to track is probably server-validated shooting an actor on another server (so Input > Server RPC > Auth Server RPC > Client RPC). Our Actor::TakeDamage is cross-server by default so likely something that will be present in zoned games (and the same case for property updates)
> 
> It also gives us an easy base to compare with single-worker too which is really handy as it introduces an additional runtime/worker hop
> 
> Thoughts? cc @m-samiec @MatthewSandfordImprobable

I guess it depends what we are going for. Are we looking to simulate the flow of a real game or do we want to isolate an measure individual parts of the flow. This seems to be more on the "measure individual parts" side but is not perfect. I think perhaps you are right Danny and that we should use event tracing to work out timing breakdown.

m-samiec:
@jayprobable is there any logic to make sure servers don't pick actors that might migrate? Perhaps worth configuring this test so that the number of migrating actors is 0 (instead of 25%)

m-samiec:
@jayprobable just looking at the LatencyGameMode changes - where does the ClientTraceMetadata get set?

m-samiec:
The looping variable that is passed into the SetTimerByEvent function is only evaluated when the event is initially setup, not each time the event runs. So I think we need to always loop, but cancel the event on authority lost.

m-samiec:
Why is there a distance check in the `Find Target in Other Zone` check?

m-samiec:
For `OnInput_CrossServerCall` to be valid, the actor we're attempting to execute the reply on shouldn't be authoritative locally, worth adding some validate/error checking in that function

m-samiec:
I'm a little concerned that the `Fine Target in Other Zone` won't play nicely with the RPC > Property trace, because the order the that the server iterates over actors could be the same each time. This means a server might attempt to reuse the replicated property multiple times before a single trace is complete (potentially true for RPC > RPC also, but as they are transient it's less critical). Can the server have some logic so that it caches the actors it picked recently, and doesn't repick within a certain threshold?

jayprobable:
> The looping variable that is passed into the SetTimerByEvent function is only evaluated when the event is initially setup, not each time the event runs. So I think we need to always loop, but cancel the event on authority lost.

Yes, I've reviewed some documentation, I will manage this by storing the handle and canceling the timer.

jayprobable:
> I'm a little concerned that the `Fine Target in Other Zone` won't play nicely with the RPC > Property trace, because the order the that the server iterates over actors could be the same each time. This means a server might attempt to reuse the replicated property multiple times before a single trace is complete (potentially true for RPC > RPC also, but as they are transient it's less critical). Can the server have some logic so that it caches the actors it picked recently, and doesn't repick within a certain threshold?

I've added code to track which actors are "InUse" for tracing, this invovles maintaining an array of "targets" in the game mode and some calls to add/remove. I doubt this is fool proof, during the course of the previous sprint I tried various approaches to this problem and have concluded this is reasonable, especially given that it's working with constraints specific to how tracing RPCs works and doesn't represent real game play.

Here's the nfr framework test run: https://buildkite.com/improbable/product-research-benchmarks-run/builds/16383

jayprobable:
Also just to be clear, the reason we're starting this trace on the server is that I'm unaware of another way to find a target on "another" worker. So the basic case here is find targets over which we are not authoritative. Please let me know if there would be a simple way to find such targets from the client? Then I could easily switch to client -> server -> auth server -> client...

The distance check is because we're not selecting the target from the client where net cull distance takes effect, using this check effectively prevents choosing a target "out of range", based on the results since adding the check it appears to be working well.

Regarding the migrating actors, I've been presuming we want some amount of migration as part of this scenario, while it is possible that an actor migrates this just results in that specific trace not being included. So far from the test results this does not appear to be a significant issue and I believe it's best we leave the scenario as is.

PR #206, opened 2020-09-27
UNR-4288 Fix DropCube increasing issue

yujunting:
When running zoning-nfr on buildkite or cloud deployment, We noticed an issue about the count of DropCube is more than expectation(maybe more than 1000).
After analyzing the logs of cloud deployment, we found the reason is in SpawnDestoryCube_BP.
After Delay behind SpawnDropcube, the player(or npc) maybe lost authority by migrating to another worker. So that the  Set ShouldSelfDestuct failed without authority.
In order to fix this issue, we do something like this:
1) Add Replicate EventGraph OnSetShouldSelfDestruct in DropCube_BP
2) Call OnSetShouldSelfDestruct when the player(or npc) has no authority in SpawnDestroyCube_BP
3) Use GetSpawnedCube after Delay and add IsValid and HasAuthority check before call OnSetShouldSelfDestruct 
4) Change the InterestBorder to 15000cm in BP_BenchmarkGym_Zoning2_1(and other BP_BenchmarkGym_Zoning(3,4,5)_1)

Jira ticket:https://improbableio.atlassian.net/browse/UNR-4288

before changes:https://buildkite.com/improbable/unrealgdk-nfr/builds/2843
after changes(compare with native, 1-worker, 2-workers):https://buildkite.com/improbable/unrealgdk-nfr/builds/2855
BluePrint ScreenShot:
image


MatthewSandfordImprobable (approved):

MatthewSandfordImprobable:
Accidently merged this branch :/

Danny perhaps you can give this a look over as well?

improbable-danny:
Confused about OnAuthorityLoss to call OnSetSelfDestruct, does the delay not call this in the case of authority/non-authority?

PR #208, opened 2020-10-02
Update plane size in SpatialNetworkingMap

eviK88:
Increase the plane size in SpatialNetworkingMap to enable VisibilityTest changes


mironec (approved):
Seems good

PR #209, opened 2020-10-05
.umap for property replication test

oblm:
This PR contains a .umap that https://github.com/spatialos/UnrealGDK/pull/2594 depends on.


oblm:
This `.umap` was opened in UnrealEngine 4.25.x, meaning that it's not compatible with `4.24.x` versions of the Unreal Engine.

This is a problem as we support both UnrealEngine versions, however, I don't think fixing this problem is in scope of this PR. We need a systematic solution to this.

mironec:
I have briefly looked at a systematic solution for this, and unfortunately it seems REALLY difficult to do (as in a few weeks of engineering effort). I think for now, it may be easiest to just get someone with 4.24 (e.g. me) to reconstruct the map.

oblm:
> I have briefly looked at a systematic solution for this, and unfortunately it seems REALLY difficult to do (as in a few weeks of engineering effort). I think for now, it may be easiest to just get someone with 4.24 (e.g. me) to reconstruct the map.

In that case yes I think reconstructing the map is our best bet, please do that.

Separately, I will ensure that "write and save tests in 4.24" is documented in the maintainers guide.

oblm:
Closing this as we now generate maps from C++ at runtime in order to ensure our tests are compatible with multiple engine versions.

oblm:
This PR was replaced by: https://github.com/spatialos/UnrealGDKTestGyms/pull/318

PR #211, opened 2020-10-08
Add config ui input mode test gym

KeyMaster-:
This gym tests that you can register the "OnConfigUIClosed" callback on the spatial debugger.
It also gives an example of using the callback to return your game to the correct input mode depending on whether the game has a UI open or not.

Input mode switching demo:
config_ui_test_gyms_demo

Blueprints:
Map:
2020-10-08 15_57_53-GDKTestGyms - Unreal Editor
Player controller implementation:
2020-10-08 15_59_26-GDKTestGyms - Unreal Editor
The player controller sets the game input mode, which is why the callback is implemented there.


m-samiec (approved):
Nice - please add a section to the `USER_MANUAL.md` also

PR #213, opened 2020-10-09
GDK Test Gym updates for the task to add worker information to the boundaries v4.24 compatible

Vix2021:
**Description**
- Increase the height of the worker boundary so the text is visible. 
- Add a small quad load balancing strategy and apply to the Handover map.
- Cook Spatial Debugger folder for materials and fonts to be in the cloud deployments. 
[Copied changes from 4.25 to 4.24 for backward compatibility]

image

**Asset changes**
BP_VerboseSpatialDebugger (minor change): increase the vertical scale of the worker region so that the text is visible
image
BP_QuadrantSmallLBStrategy (new asset)
image
BP_QuadrantSmallZoningSettings (new asset)
image
HandoverGym (minor change): Change the multi worker class settings to the new BP_QuadrantSmallZoningSettings 
image

**Test**
Locally with the Handover gym and Teleport gym and in the cloud with the Handover gym to check that the worker boundaries are visible and the information displayed correctly.
Cloud deploy: 

**Related PR**
https://github.com/spatialos/UnrealGDK/pull/2606


Game/Config/DefaultGame.ini
```diff
@@ -5,6 +5,7 @@ bSpatialNetworking=True
 
 [/Script/UnrealEd.ProjectPackagingSettings]
 +DirectoriesToAlwaysCook=(Path="Spatial")
++DirectoriesToAlwaysCook=(Path="/SpatialGDK/SpatialDebugger/")
```
m-samiec, line 8: As discussed offline - think this should also be done as an engine change - can be complete as a follow up

m-samiec (approved):

PR #218, opened 2020-10-21
UNR-4276 Workaround no longer required

improbable-danny:
This spawning issue was fixed in the GDK, a stable reference was being passed to a non-replicated actor but now a null reference is passed and the spawn flow completes correctly (https://github.com/spatialos/UnrealGDK/pull/2592)


yujunting (approved):
Nice job, you did what I wanted to do but wasn't sure I could do.
Find out the DefaultPlayerStart as the playerstart for all the players is not safe for NFR.

PR #222, opened 2020-11-02
UNR-4426 fix connected clients expectation synchronization issue

yujunting:
Sometimes, there are some 1 or 2 clients lost in the max-player scenario of zoning-NFR. 
After investigating the client_stayed_connection issue. I found the value of total players on all the workers is not a stable value.
As the picture is shown on the [ticket](https://improbableio.atlassian.net/browse/UNR-4426).
So we decide to change the total player number as an average value for the last frame and the current frame.
It should very approach the value we expected. 
Sometimes it was 1.xx less than expectation. So we set the expectation of max-player as 2.

BTW:
1. Change the player number check to a valid check. So we can have the same value of expectation with BuildKite.
2. Change the timer for the delay, check and report. to make sure when we start to check, the worker has enough time to prepare the simulate players and reduce the frequency of RPC of when report.
3. Change the migration to total migration, not for a single worker.

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -23,7 +23,7 @@ namespace
 {
 	const FString AverageClientRTTMetricName = TEXT("UnrealAverageClientRTT");
 	const FString AverageClientUpdateTimeDeltaMetricName = TEXT("UnrealAverageClientUpdateTimeDelta");
-	const FString PlayersSpawnedMetricName = TEXT("UnrealActivePlayers");
+	const FString ExpectedPlayersValidMetricName = TEXT("ExpectedPlayersValid");
 	const FString ActorMigrationValidMetricName = TEXT("UnrealActorMigration");
 	const FString AverageFPSValid = TEXT("UnrealServerFPSValid");
 	const FString AverageClientFPSValid = TEXT("UnrealClientFPSValid");
@@ -46,6 +46,9 @@ namespace
 
 	const FString MinActorMigrationWorkerFlag = TEXT("min_actor_migration");
 	const FString MinActorMigrationCommandLineKey = TEXT("-MinActorMigration=");
+	
+	const FString NumWorkersWorkerFlag = TEXT("num_workers");
+	const FString NumWorkersCommandLineKey = TEXT("-NumWorkers=");
 
 	const FString NFRFailureString = TEXT("NFR scenario failed");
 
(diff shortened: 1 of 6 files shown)


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -618,41 +644,44 @@ void ABenchmarkGymGameModeBase::TickActorMigration(float DeltaSeconds)
...
+				// Only report AverageMigrationOfCurrentWorkerPerSecond to the worker which has authority
+				float AverageMigrationOfCurrentWorkerPerSecond = MigrationOfCurrentWorker / MigrationSeconds;
+				ReportMigration(FPlatformProcess::ComputerName(), AverageMigrationOfCurrentWorkerPerSecond);
+				ActorMigrationReportTimer.SetTimer(1);
+			}
 
 			if (HasAuthority() && ActorMigrationCheckTimer.HasTimerGoneOff())
 			{
-				float TotalMigrations = 0;
+				float Migration = 0;
```
MatthewSandfordImprobable, line 657: Nit:   float Migration = 0.0f;

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -578,10 +594,23 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayers_Implementation(const
 	if (HasAuthority())
 	{
 		MapAuthoritativePlayers.Emplace(WorkerID, AuthoritativePlayers);
-		ActivePlayers = 0;
+		int TotalPlayers = 0;
```
MatthewSandfordImprobable, line 597: Super Nit: int32t

MatthewSandfordImprobable (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -578,10 +594,23 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayers_Implementation(const
...
+		}
+		if (MapAuthoritativePlayers.Num() == NumWorkers)
+		{
+			if (AveragedTotalAuthPlayers < 0.0f)
+			{
+				AveragedTotalAuthPlayers = TotalPlayers;
+			}
+			else
+			{
+				AveragedTotalAuthPlayers = (AveragedTotalAuthPlayers + TotalPlayers) * 0.5f;
```
improbable-danny, line 610: SmoothedTotalAuthPlayers might make more sense here, and maybe using the formula  `SmoothedTotalAuthPlayers*(1 - x) + TotalPlayers*x` (here x = 0.5) is slightly more intuitive with what is being done

improbable-danny (approved):
1 comment otherwise looks good 👍

jayprobable (approved):
Generally speaking don't combine different tasks into one PR, e.g. the changes for migration could have been done in a separate PR.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
yujunting: seems good to change the x to 0.1 to make the value more smoothed and can show the real loss of clients
SmoothedTotalAuthPlayers*(1 - x) + TotalPlayers*x

PR #232, opened 2020-11-11
[UNR-4449] Add a small delay to allow replication to occur in attack test

improbable-danny:
This step needs to wait for for the other client's attack RPC / property update

image


improbable-valentyn (changes requested):
Still fails with the same error.

PR #236, opened 2020-11-19
bugfix/UNR-3289/Fix spatial commands on workers on native windows

antoinecordelle:
When working on native windows (such as cmd), echo wasn't found in the path and the commands failed. 

This PR ensures the modified commands work on native windows while still working using unix toolings (cmdler for instance) on Windows


simonsarginson:
Does this not have any side effects (as mentioned it could possibly have in the ticket?)

antoinecordelle:
This solution is a trade-off between Mac and Window:
- Will work anyway: Cmdler, Git Bash and other Unix tooling based terminals
- Worked and won't work anymore: MacOS 
- Will now work: Powershell and Windows cmd prompt

```"steps": [{"name": "No-op", "command": "echo", "arguments": ["No-op."]}]```
The command field directly calls the specified key in the path, so we can't inject any conditional statement on the OS here (or anything that is not on the path, so for instance even `cmd /k echo` won't work as it's not in the path). The json tree used has a rigid structure and it doesn't seem to accept `windows` or `macos` conditions on the build commands.

jessicafalk (approved):
As windows is our priority and these print statements are only required to tell people that this is a no-op (they should use the Unreal Editor instead of invoking the command directly), this lgtm!

improbable-danny (approved):

PR #238, opened 2020-11-20
[UNR-4474] Remote ability activation testing in Ability Locking Gym

KeyMaster-:
Updates the ability locking gym to use the remote ability activation functions added in https://github.com/improbableio/UnrealEngine/pull/692. All blueprint-accessible variants of abiility activation (by class/tag/event) are exposed.
Also adds a way to attempt ability activation without using the cross-server methods, to test that an appropriate warning is emitted when the activation fails due to lack of authority.

Renamed to Ability Activation Gym

#### AbilityLockingGym
- Changed the instruction text.
2020-11-24 11_41_44-GDKTestGyms - Unreal Editor

#### AbilityLockingController
- Reworked most client commands to send a single client->server RPC and find all actors to affect on the server.
- Replaced any uses of the counting cube's "ActivateAbility" cross-server RPC with "TryActivateAbilityByClassCrossServer".
- Added commands to activate the ability by event or tag.
2020-11-24 12_27_08-GDKTestGyms - Unreal Editor
2020-11-24 12_27_12-GDKTestGyms - Unreal Editor
- Added command to activate the abilty without using the cross-server functions, to test that a warning is logged in this case. 
2020-11-24 12_23_31-GDKTestGyms - Unreal Editor

#### CountingCube:
- Reparented to new GasCube native class.
- Duplicated the functionality of previous parent class (GymCube) and merged it directly into the CountingCube blueprint class.
The GymCube functionality was what made the cube move back and forth across a boundary.
- Removed the "ActivateAbility" cross-server RPC that activates the counting ability.

#### Ability_CountingFromEvent
- Mostly a duplicate of Ability_Counting, but without the "GameplayAbility.Counting" tag as an ability tag.
- Added an ability trigger with trigger tag "GameplayAbility.Trigger" and trigger source "Gameplay Event".
- Replaced the ActivateAbility node with ActivateAbilityFromEvent (otherwise this abiility does not activate from events).
- Added code to print the tag with which the ability was activated at the start of ActivateAbilityFromEvent.
2020-11-24 12_22_54-GDKTestGyms - Unreal Editor

Effect_GrantCounting
- Added Ability_CountingFromEvent to the list of granted abilities, such that either can be activated on the counting cube.


Game/Source/GDKTestGyms/Private/GasCube.cpp
```diff
@@ -0,0 +1,11 @@
...
+
+#include "GasCube.h"
+
+AGasCube::AGasCube()
+{
+	bReplicates = true;
+
+	AbilitySystemComponent = CreateDefaultSubobject<UAbilitySystemComponent>(TEXT("AbilitySystemComponent"));
+	AbilitySystemComponent->SetIsReplicated(true);
+}
```
simonsarginson, line 11: minor: missing newline at end of file.

simonsarginson (approved):

m-samiec:
What lbs does this map use now?

m-samiec:
> Duplicated the functionality of previous parent class (GymCube) and merged it directly into the CountingCube blueprint class.
> The GymCube functionality was what made the cube move back and forth across a boundary.
Why duplicate the functionality?

m-samiec:
> TODO: This gym needs a better name now. Suggestions welcome.

Looks like you've settled on Ability Activation Gym - sounds good to me - have all the naming updates been made? (I see in the screen shot it's still ability locking gym)

Game/Source/GDKTestGyms/Private/GasCube.cpp
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Improbable Worlds Ltd, All Rights Reserved
+
+#include "GasCube.h"
+
+AGasCube::AGasCube()
```
m-samiec, line 5: This cube isn't a cube at all! Where's the static mesh?

USER_MANUAL.md
```diff
@@ -82,19 +82,29 @@ The ReplicatedStartupActorTest is failing, pending https://improbableio.atlassia
...
   * A player ownership hierarchy of Actors (controller, character & state) will not migrate while one Actor in the hierarchy is locked.
 * Contains:
   * a cube moving across a server boundary, with a gameplay ability granted to it which takes 4 seconds to complete.
   * a player with a gameplay ability granted to it which takes 4 seconds to complete.
 * To test the gym:
   * If it is working correctly the authority and authority intent of the cube can be seen to change as it moves across the floor, and the text "Uninitialized" hovering over the cube.
-  * Press "Q" to start the ability on the player.
-  * Press "T" to start the ability on the cube.
+  * Activate an ability on the player or the cube:
+    * Press "Q" to activate the ability on the player.
```
m-samiec, line 96: Could you please give a quick description of what the user should expect to see if these key presses succeed/fail

m-samiec:
Few comments - looking good

KeyMaster-:
> 
> 
> What lbs does this map use now?

Same as before, the 4-worker strategy.



> 
> 
> > Duplicated the functionality of previous parent class (GymCube) and merged it directly into the CountingCube blueprint class.
> > The GymCube functionality was what made the cube move back and forth across a boundary.
> 
> Why duplicate the functionality?

The ASC needs to be added at the C++ level. Previously the inheritance chain for CountingCube was CountingCube -> GymCube -> Actor (child -> parent). To add the ASC I'd have to give GymCube a C++ parent with an ASC, but that means that all children of GymCube get the ASC, which I don't think they should. So instead I gave CountingCube its own C++ parent class, but that means it can't inherit from GymCube anymore, so it gets the functionality from GymCube copied into it.
Alternatively we could have both the current GymCube and a new GasGymCube, with the required C++ parent. I just didn't think that was worth doing for what's currently a single asset that would inherit from that class.


> 
> 
> > TODO: This gym needs a better name now. Suggestions welcome.
> 
> Looks like you've settled on Ability Activation Gym - sounds good to me - have all the naming updates been made? (I see in the screen shot it's still ability locking gym)

Good spot - I didn't update the screenshots, but I can confirm all the assets have been renamed.

Game/Source/GDKTestGyms/Private/GasCube.cpp
KeyMaster-: Ehh good point. The static mesh currently lives in the blueprint asset, which I think is the right place. Will rename this to just "AGasActor", as it is just that. (Also the doc comment for the class is wrong, will also fix)

USER_MANUAL.md
KeyMaster-: The text below the list of inputs describes the results of activating an ability - the intent here is that all of the different key presses activate the same ability and thus have the same visual result.
I've clarified the description of what success looks like, lmk if you think this still needs to be made clearer.

m-samiec (approved):

PR #240, opened 2020-11-24
Change the attack initiation to also check the owner is set

improbable-danny:
This test would sometimes flake because the owner was not set on the client, which is a requirement for being able call RPCs on it. V15 also fixed an issue where the client ownership is tied to having spatial authority 🎉 

Run 20/20 without failure

Changed Client_1_Attack to a repeated tick step and also check bIsClientOwned and the owner is set correctly
image


ImprobableNic:
Didn't we want to fix it with checking IsClientOwner ?

MatthewSandfordImprobable:
Just so I understand how is this stopping the flake? Previously if owner was null I guess we would fail to send an RPC. Now we just will simply not attempt to send the RPC if owner is null. I guess something has changed underlying this (perhaps the addition of v15)?

improbable-danny:
>Didn't we want to fix it with checking IsClientOwner ? 

Yeah, although this is no longer required because of the authority changes in v15! Had a chat with Mike about it and there was potentially another way to fix it in the RPC service for clients to not use the data component (and append to the update for when authority is gained - may still be worth doing). The owner and Unreal authority should align with spatial authority now though 

>Just so I understand how is this stopping the flake? Previously if owner was null I guess we would fail to send an RPC. Now we just will simply not attempt to send the RPC if owner is null. I guess something has changed underlying this (perhaps the addition of v15)?

There is a race where the client runs before the target actor being owned by the clients flow controller - it was quite rare ~2/20 but this has resolved that. This change also modified the step to also run on tick so it will try to run again until the attacker is valid and has a valid owner

ImprobableNic:
>The owner and Unreal authority should align with spatial authority now though
AFAIK, the load balancing is still decentralized, so I wonder how that is possible.

> (and append to the update for when authority is gained - may still be worth doing)
I'd rather have an error in that case, because it might be more likely that the RPC was sent by mistake.

improbable-danny:
It does actually still reproduce with just IsClientOwned - maybe it's worth using IsOwner() != null && IsClientOwned() to ensure the attacker is valid

ImprobableNic:
Wow this is weird. IsClientOwned == true should imply that Owner == Client's controller. Unless we are able to get an update on auth faster than the update that caused it to switch in the first place? Seems wrong.

improbable-danny:
I think we can track that issue in UNR-4503, for now I'll just fix the test to unblock 4.26. Seems to be a lot deeper and complex than originally thought!

MatthewSandfordImprobable (approved):

m-samiec (approved):

PR #242, opened 2020-12-01
UNR-3519 - Add ClientTravel Test Gym

antoinecordelle:
Add a ClientTravel test gym.

This requires to create a Cloud deployment in `unreal_gdk_starter_project` named `client_travel_gym`, and then press K to trigger a ClientTravel for the PlayerController to the same deployment.

Linked to https://github.com/spatialos/UnrealGDK/pull/2719


joshuahuburn (approved):
Can you perhaps show some screenshots of the BPs?

joshuahuburn (commented):
meant to make a comment, not approve :P

antoinecordelle:
Sure! 

The Gamemode only specifies that the PlayerController used should be a ClientTravelPlayerController_BP
Here is the ClientTravelPlayerController_BP: it is a BP class inheriting from the C++ ClientTravelPlayerController (There might be better ways of implementing that, I'm not familiar with Blueprints)
image

jessicafalk (approved):

PR #243, opened 2020-12-02
Feature/add multiworker world comp

m-samiec:
Test gym for https://github.com/improbableio/UnrealEngine/pull/706

Add multiworker world composition gym with no interest borders to ensure that always relevant actors are replicated to interested clients even when those auth servers don't have the player controllers checked out.

Multiworker world composition gym has two sublevels, each with a single NoNetLoadGymCube. 
NoNetLoadGymCube extends from GymCube, and just has bNetLoadOnNonAuthClient set to false.
The cubes movement waypoints are setup to move from one server to the other.

Drivebys
TeleportingPlayerController.uasset reverts `bReplicates` being set to true, to allow it to function with the replication graph. PlayerControllers by default don't have `bReplicates` set to true and are considered for replication via their NetController.
Clang format pass on the TestGymsReplicationGraph.cpp

image


USER_MANUAL.md
```diff
@@ -306,5 +306,11 @@ The ReplicatedStartupActorTest is failing, pending https://improbableio.atlassia
   * Launch the game in PIE mode, connecting to the corresponding Cloud Deployment.
   * Press K to trigger a ClientTravel for the PlayerController to the same deployment. It should be visible in PIE as your position in the map will be reset.
 
+##### Multiworker World Composition gym
```
ImprobableNic, line 309: Should we add that the RepGraph should be enabled for this gym ?

ImprobableNic (approved):
LGTM, minor nit

PR #255, opened 2020-12-18
UNR-988 added functional test map  for testing zombie connection bug

simonsarginson:
Added functional test map and worker setting for testing zombie connection bug.
https://github.com/spatialos/UnrealGDK/pull/2779


mironec:
Ah, not sure if we have already set precedent for this, but can we define the settings and strategy in C++?

simonsarginson:
@mironec we set the multi worker settings class on the map directly, so it would be a painful to define the strategy if we don't generate the map at runtime (which we don't). I think it would require a string name lookup of the test map which seems pretty fragile to me.

mironec:
Ah, sorry, I was unclear, I thought that e.g. `Game/Content/FunctionalTestSamples/BP_1x2_SmallInterest_WorkerSettings.uasset` could instead be a `1x2_SmallInterest_WorkerSettings.cpp`, similarly with `Game/Content/FunctionalTestSamples/Test1x2SmallInterestGridStrategy.uasset`. But I am not sure what our other tests have followed.

mironec (approved):

improbable-valentyn (approved):

PR #266, opened 2021-01-11
Tell test users to generate schema and snapshot.

oblm:
Tell test users to generate schema and snapshot.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -121,31 +121,25 @@ The ReplicatedStartupActorTest is failing, pending https://improbableio.atlassia
 * To see results of the tests, go to https://console.cloud.google.com/traces/traces?project=holocentroid-aimful-6523579
 
 ##### Unresolved reference gym
-* Test what happens when structs with references to actors whose entity have not been created yet are replicated. Replicating null references is accepted, but they should be resolved eventually.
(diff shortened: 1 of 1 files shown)


mironec (approved):
I approve, just wondering if we want to update all the other instructions in this file, e.g.:
```
##### Net reference test gym
* NOTE: This gym also has an equivalent automated test. In order to run the test, follow the steps:
  1. Open the Session Frontend: Window -> Developer Tools -> Session Frontend.
  2. On the Automation tab, search for SpatialTestNetReference1, tick the box corresponding to it and hit Start Tests.
  3. The Session Frontend will then prompt you with the result of the test.
  * Test that references to replicated actors are stable when actors go in and out of relevance
```
I would prefer removing these instructions and referring to these instead.

oblm:
> I approve, just wondering if we want to update all the other instructions in this file, e.g.:
> 
> ```
> ##### Net reference test gym
> * NOTE: This gym also has an equivalent automated test. In order to run the test, follow the steps:
>   1. Open the Session Frontend: Window -> Developer Tools -> Session Frontend.
>   2. On the Automation tab, search for SpatialTestNetReference1, tick the box corresponding to it and hit Start Tests.
>   3. The Session Frontend will then prompt you with the result of the test.
>   * Test that references to replicated actors are stable when actors go in and out of relevance
> ```
> 
> I would prefer removing these instructions and referring to the ones above instead.

Good idea. Done.

PR #268, opened 2021-01-12
Recreation of SpatialZoningMap in 4.24

KeyMaster-:
Also recreates the appropriate worker settings it uses along with it.


mironec:
I like it, but since we are here.........
`Game/Content/FunctionalTestSamples/BP_1x2_WorkerSettings.uasset` can de done in C++, I think. Just in case you wanted to. Not necessary. Would work in all engine versions then. But not needed. We could even use macros if something broke. But just a random suggestion, approving anyway.

mironec (approved):

KeyMaster-:
The LB strat is already defined in C++, so I might as well have a look at doing the worker settings as well.

KeyMaster-:
Yep that was easy enough to do, see https://github.com/spatialos/UnrealGDK/pull/2811 for the PR of the C++ settings class.

PR #279, opened 2021-01-18
[UNR-4827] Move GAS tests into GDKTestGyms

KeyMaster-:
Moves the two tests and their test actor dependencies.
The tests were re-added to SpatialZoningMap, with the PredictedGameplayCuesTest disabled since it currently fails in CI for unknown reasons.


Game/GDKTestGyms.uproject
```diff
@@ -14,6 +14,10 @@
 		}
 	],
 	"Plugins": [
+		{
+			"Name": "GameplayAbilities",
```
KeyMaster-, line 18: I'm not sure if this is needed - the UnrealGDK plugin isn't listed here either but still used.

Game/Source/GDKTestGyms/FunctionalTests/TestActors/GASTestPawnBase.h
```diff
@@ -0,0 +1,41 @@
...
+#include "AbilitySystemInterface.h"
+#include "CoreMinimal.h"
+#include "GameFramework/Pawn.h"
+#include "GASTestPawnBase.generated.h"
+
+/**
+ * A replicated Actor with a Cube Mesh, used as a base for Actors used in spatial tests.
+ */
+UCLASS()
+class AGASTestPawnBase : public APawn, public IAbilitySystemInterface
```
KeyMaster-, line 16: There's a fair bit of code duplication between this and the GASCharacter class that already exists in the test gyms project - the only difference is that this class has the automaticlaly granted abilities. 
This class could feasibly be a subclass of the existing one, but I'm not sure if we want to mix gyms and test classes.

Game/Source/GDKTestGyms/FunctionalTests/TestActors/GASTestActorBase.h
```diff
@@ -0,0 +1,38 @@
...
+#include "AbilitySystemComponent.h"
+#include "AbilitySystemInterface.h"
+#include "CoreMinimal.h"
+#include "GASTestActorBase.generated.h"
+
+/**
+ * A replicated Actor with a Cube Mesh, used as a base for Actors used in spatial tests.
+ */
+UCLASS()
+class AGASTestActorBase : public AActor, public IAbilitySystemInterface
```
KeyMaster-, line 15: Same code duplication concerns as for AGASTestPawnBase

MatthewSandfordImprobable:
Concerning the code duplication, I don't think I am too fussed. In my opinion, I think it is marginally better to keep them separate.

MatthewSandfordImprobable (approved):

PR #285, opened 2021-01-19
Fixing flake in NFRs to do with player count

MatthewSandfordImprobable:
Using "NumSpawnZones" to distribute spawn clusters. This is just to ensure we have parity between offloaded, non offloaded and native tests.

NFRBuildkite PR: https://github.com/improbable/TestGymBuildkite/pull/219

BK Test: https://buildkite.com/improbable/unrealgdk-nfr/builds/3544

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -49,6 +49,9 @@ namespace
 	
 	const FString NumWorkersWorkerFlag = TEXT("num_workers");
 	const FString NumWorkersCommandLineKey = TEXT("-NumWorkers=");
+
+	const FString NumSpawnZonesWorkerFlag = TEXT("num_spawn_zones");
+	const FString NumSpawnZonesCommandLineKey = TEXT("-NumSpawnZones=");
 #if	STATS
 	const FString StatProfileWorkerFlag = TEXT("stat_profile");
 	const FString StatProfileCommandLineKey = TEXT("-StatProfile=");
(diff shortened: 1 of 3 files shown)


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -49,6 +49,9 @@ namespace
 	
 	const FString NumWorkersWorkerFlag = TEXT("num_workers");
 	const FString NumWorkersCommandLineKey = TEXT("-NumWorkers=");
+
+	const FString NumSpawnZonesWorkerFlag = TEXT("num_spawn_zones");
+	const FString NumSpawnZonesCommandLineKey = TEXT("-NumSpawnZones=");
```
improbable-danny, line 54: Maybe just NumZones would be enough? We don't use that terminology for offloading really

improbable-danny (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Well this was just to separate it from multi worker completely. Now you just ask it to change how the spawn clusters are spawned. If there is a better way to encapsulate this I am happy to change it. NumZones ties it into multi worker IMO.

m-samiec (approved):

m-samiec:
When this goes in can we update master from the rc also so we can see this fix in the nightlies?

MatthewSandfordImprobable:
> When this goes in can we update master from the rc also so we can see this fix in the nightlies?

Sure!

PR #286, opened 2021-01-20
Update Cross-Server Possession gym to USER_MANUAL.md and PossessionGym

yujunting:
Add user-manual for RemotePossessionComponent functional tests.


USER_MANUAL.md
```diff
@@ -336,6 +336,16 @@ These tests can only be run automatically. To run them:
...
   * Press K to trigger a ClientTravel for the PlayerController to the same deployment. It should be visible in PIE as your position in the map will be reset.
+  
+##### Cross-Server possession gym
+* Functional tests the Controller remote possess the Pawn
+* How to test:
+  * Set the current map to `PossessionGym`
+  * Set the `GameMode Override` to `CrossServerPossessionGameMode`(World Settings > Game Mode)
+  * Set the `Multi Worker Settings Class` to `BP_Possession_Settings_Zoning2_2`(World Settings > Multi Worker)
+  * Run functional tests such as `CrossServerMultiPossessionTest1`, `CrossServerPossessionLockTest1`, `CrossServerPossessionTest1`, `NoneCrossServerPossessionTest1` (Window > Developer Tools > Session Frontend > Automation > Project > Functional Tests > /Game/Maps/PossessionGym(4))
+  * NOTE: Do not run all tests together
```
m-samiec, line 347: Why is this?

USER_MANUAL.md
```diff
@@ -336,6 +336,16 @@ These tests can only be run automatically. To run them:
   * Create a Cloud deployment in `unreal_gdk_starter_project` named `client_travel_gym`.
   * Launch the game in PIE mode, connecting to the corresponding Cloud Deployment.
   * Press K to trigger a ClientTravel for the PlayerController to the same deployment. It should be visible in PIE as your position in the map will be reset.
+  
+##### Cross-Server possession gym
+* Functional tests the Controller remote possess the Pawn
+* How to test:
+  * Set the current map to `PossessionGym`
+  * Set the `GameMode Override` to `CrossServerPossessionGameMode`(World Settings > Game Mode)
```
m-samiec, line 344: What content is being reused between the possession tests?
I think we should create a separate map, so the user doesn't need to change anything.

USER_MANUAL.md
yujunting: Related to the second NOTE:Set Num Required Clients of CrossServerMultiPossessionTest1 to 1 Before run NoneCrossServerPossessionTest1.
Our functional test decides the client number as the max one of the map.
The CrossServerMultiPossessionTest1 should have more than 1 client but NoneCrossServerPossessionTest1 should have only 1 client.

USER_MANUAL.md
yujunting: Good idea!

yujunting: After refactoring, those 2 NOTEs can be removed

PR #297, opened 2021-01-25
Clarifies `Client Travel gym` test steps.

oblm:
* Clarifies `Client Travel gym` test steps.
* I ran these steps to validate that they are correct.
* I rendered the markdown to ensure that I formatted it correctly.


antoinecordelle (approved):
The steps are indeed way clearer than before, thanks!

PR #299, opened 2021-01-25
[UNR-4585] Replay test gym

mironec:
Adds a test gym for testing replays.
Nothing special there, so created a few nothing special characters, since I have often found the need to have really basic characters, rather than the default NFR ones, which spawn cubes and whatnot.


Game/Source/GDKTestGymsEditor.Target.cs
```diff
@@ -10,6 +10,7 @@ public GDKTestGymsEditorTarget(TargetInfo Target) : base(Target)
 		Type = TargetType.Editor;
 		ExtraModuleNames.Add("GDKTestGyms");
 		ExtraModuleNames.Add("GDKTestGymsFunctionalTests");
+
 		DefaultBuildSettings = BuildSettingsVersion.V2;
```
mironec, line 14: Hm, I see I conflicted with changes from master here, so someone had the same idea.
Not sure if it's necessary for all targets?

Game/Source/GDKTestGyms/Public/ReplayPlayerController.h
```diff
@@ -0,0 +1,35 @@
+// Fill out your copyright notice in the Description page of Project Settings.
```
mironec, line 1: I thought I changed this default at some point...

Game/Source/GDKTestGyms/GDKTestGyms.Build.cs
```diff
@@ -19,6 +19,7 @@ public GDKTestGyms(ReadOnlyTargetRules Target) : base(Target)
 				"GameplayAbilities",
 				"GameplayTasks",
 				"InputCore",
+				"Slate",
```
mironec, line 22: Needed for the input component stuff and binding keys directly.
Maybe I could setup action mappings instead....

Game/Source/GDKTestGymsEditor.Target.cs
m-samiec: I think you're right, I should have set it in all targets

m-samiec (approved):

PR #300, opened 2021-01-28
UNR-4822 Fix TestGyms branch detection in CI

simonsarginson:
Added dynamic ci steps to build with the matching GDK_BRANCH with similar precedence rules as other GDK repos


.buildkite/premerge.definition.yaml
```diff
@@ -2,6 +2,7 @@ agent_queue_id: trigger-pipelines
 description: TestGyms premerge (calls GDK premerge)
 github:
   default_branch: master
+  pull_request_branch_filter_configuration: ["!docs*"]
```
mironec, line 5: will have to `circle ci pipeline push` after this is merged. I agree it would be nice to have a checker/auto-pusher

ci/generate-pipeline-steps.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -e -u -o pipefail
+if [[ -n "${DEBUG-}" ]]; then
+    set -x
+fi
+
+# Download the unreal-engine.version file from the GDK repo so we can run the example project builds on the same versions the GDK was run against.
+# This is not the pinnacle of engineering, as we rely on GitHub's web interface to download the file, but it seems like GitHub disallows git archive
+# which would be our other option for downloading a single file.
```
mironec, line 10: This comment is not relevant for this repo
Maybe instead I would do a comment saying this is mostly copied from EP's CI, so that we at least know when we refactor this.

ci/generate-pipeline-steps.sh
```diff
@@ -0,0 +1,29 @@
...
+# Also resolve the GDK branch to run against. The order of priority is:
```
mironec, line 11:
```suggestion
# Resolve the GDK branch to run against. The order of priority is:
```

ci/generate-pipeline-steps.sh
```diff
@@ -0,0 +1,29 @@
...
+# GDK_BRANCH envvar > same-name branch as the branch we are currently on > UnrealGDKVersion.txt > "master".
+GDK_BRANCH_LOCAL="${GDK_BRANCH:-}"
+if [ -z "${GDK_BRANCH_LOCAL}" ]; then
+    GDK_REPO_HEADS=$(git ls-remote --heads "git@github.com:spatialos/UnrealGDK.git" "${BUILDKITE_BRANCH}")
+    EXAMPLEPROJECT_REPO_HEAD="refs/heads/${BUILDKITE_BRANCH}"
```
mironec, line 16:
```suggestion
    TESTGYMS_REPO_HEAD="refs/heads/${BUILDKITE_BRANCH}"
```
and else where

ci/premerge.template.steps.yaml
```diff
@@ -0,0 +1,10 @@
+steps:
+  - label: "unreal-gdk-premerge"
+    trigger: "unrealgdk-premerge"
+    build:
+      branch: "GDK_BRANCH_PLACEHOLDER"
+      commit: "${GDK_COMMIT:-HEAD}"
+      message: "TestGyms ${BUILDKITE_BRANCH} ${BUILDKITE_MESSAGE}"
+      env:
+        TEST_REPO_BRANCH: "${BUILDKITE_BRANCH}"
+        ENGINE_NET_TEST: "true"
```
mironec, line 10: Nit: maybe an endline

mironec (approved):
I think it looks good! We could quickly test against e.g. the same-name GDK branch and just push for example something non-compiling there, so we see that CI picked it up correctly. (We should probably put annotations in CI saying which branches it's using.... but that would be a separate ticket)

.buildkite/premerge.definition.yaml
simonsarginson: done

PR #303, opened 2021-02-01
Functional tests for remote possession and update user_manual

yujunting:
1. Separate all the RemotePossession functional tests into the different maps. 
	CleanUp the test we made needs more works on migrating back and removing the Pawn. Make it in a separate map is easy to keep each of them clean.
2. Update the USER_MANUAL.md


USER_MANUAL.md
```diff
@@ -385,5 +385,12 @@ Manual steps:<br>
   1. Start the gym again and note that the level actors should be at their previous shutdown positions, not their original positions. You can do this repeatedly.
   1. The test has now passed. Don’t forget to revert the two settings changes you made before you run another test.
 
+##### Cross-Server possession gym
+* Functional tests the Controller remote possess the Pawn
+* How to test:
+  * Set the current map to the map which in the folder(/Game/Content/Maps/FunctionalTests/)
```
m-samiec, line 391: You've added 3 maps to this PR, but only added one use-case. 
If all that's required is selecting the map in the functional test framework, and running the test, there's no need to add an entry to the user_manual.

USER_MANUAL.md
```diff
@@ -385,5 +385,12 @@ Manual steps:<br>
   1. Start the gym again and note that the level actors should be at their previous shutdown positions, not their original positions. You can do this repeatedly.
   1. The test has now passed. Don’t forget to revert the two settings changes you made before you run another test.
 
+##### Cross-Server possession gym
+* Functional tests the Controller remote possess the Pawn
+* How to test:
+  * Set the current map to the map which in the folder(/Game/Content/Maps/FunctionalTests/)  
+  * Set the `Multi Worker Settings Class` to `BP_Possession_Settings_Zoning2_2`(World Settings > Multi Worker)
```
m-samiec, line 392: The point of adding new maps was so we don't have to make changes to the level setting prior to running the tests. Why is this necessary?

m-samiec (approved):

PR #307, opened 2021-02-05
[UNR-2626] Ability Giving Gym

KeyMaster-:
Adds a manual gym that tests that ability specs given to an ASC on different servers can be correctly activated via their handles.
Needs to be a manual gym instead of a functional test since the testing framework requires the "Run Under One Process" option to be enabled, and the bug that inspired this test (ability spec handles not being unique, see https://github.com/improbableio/UnrealEngine/pull/799) only occurs when running with that option turned on.


Game/Source/GDKTestGyms/Private/AbilityGivingGymCharacter.cpp
```diff
@@ -0,0 +1,23 @@
+// Fill out your copyright notice in the Description page of Project Settings.
```
oblm, line 1: Idk where this message is being generated from but please don't commit it.

USER_MANUAL.md
```diff
@@ -385,5 +385,25 @@ Manual steps:<br>
   1. Start the gym again and note that the level actors should be at their previous shutdown positions, not their original positions. You can do this repeatedly.
   1. The test has now passed. Don’t forget to revert the two settings changes you made before you run another test.
 
+##### Ability Giving Gym
+Tests that ability specs given on different servers can be activated correctly via their handle.
```
oblm, line 389:
```suggestion
Tests that ability specifications given to an AbilitySystemComponent on two different servers can be activated correctly via their handles.
```

USER_MANUAL.md
```diff
@@ -385,5 +385,25 @@ Manual steps:<br>
...
+* How to test:
+  * Go to `Edit > Editor Preferences > Level Editor - Play > Multiplayer Options > Run Under One Process`. Disable this option.
+  * Play with one client.
+  * On the server that you spawned on, press Q.
```
oblm, line 393:
```suggestion
  * In the client, with your character still on the server that it spawned on, press `Q` on your keyboard.
```

USER_MANUAL.md
```diff
@@ -385,5 +385,25 @@ Manual steps:<br>
...
+    In the spatial output log, you should see the following two lines:
+    > Giving and running ability with level 1
+    > 
+    > Ability activated on AbilityGivingGymCharacter_BP with Level 1
+    
+    Importantly, the level number stated in the two lines should match. 
+  * Move the character to the other server, press E.
```
oblm, line 400:
```suggestion
  * Move the character to the other server and then press `E` on your keyboard.
```

USER_MANUAL.md
```diff
@@ -385,5 +385,25 @@ Manual steps:<br>
...
+    In the spatial output log, you should see the following two lines:
+    > Giving and running ability with level 2
+    >
+    > Ability activated on AbilityGivingGymCharacter_BP with Level 2
+    
+    Again, the two level numbers should match.
```
oblm, line 406:
```suggestion
    Again, the two level numbers should match. If they do, the test has passed.
```

oblm (approved):
Made some small tweaks to the language in the user manual but otherwise lgtm.

Don't forget to create a ticket for:
>Be able to run functional tests under multiple processes

as this should not really block out ability to run this as an automated functional test.

USER_MANUAL.md
KeyMaster-: Calling them specs here because the type name in the code is `FGameplayAbilitySpec`. Do you feel strongly about the written out word?

oblm: No, in that case spec is fine. I'd still prefer that you include the rest of the changes in this snippet.

KeyMaster-: Yea ofc

Game/Source/GDKTestGyms/Private/AbilityGivingGymCharacter.cpp
```diff
@@ -0,0 +1,23 @@
...
+	Super::SetupPlayerInputComponent(PlayerInputComponent);
+
+	check(PlayerInputComponent);
+	PlayerInputComponent->BindKey(EKeys::Q, EInputEvent::IE_Pressed, this, &AAbilityGivingGymCharacter::GiveAbilityOne);
+	PlayerInputComponent->BindKey(EKeys::E, EInputEvent::IE_Pressed, this, &AAbilityGivingGymCharacter::GiveAbilityTwo);
+}
+
+void AAbilityGivingGymCharacter::ServerGiveAbility_Implementation(int32 Level)
+{
+	UE_LOG(LogTemp, Log, TEXT("Giving and running ability with level %d"), Level);
```
improbable-dmitrii, line 19: Log category change maybe?

PR #308, opened 2021-02-07
Feature/add async package loading gym

m-samiec:
Added async package loading gym to validate async loading behaviour.

@jacquese @samiwh


Game/Source/GDKTestGyms/Private/AsyncPlayerController.cpp
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		if (bAsyncTestCorrectlySetup && bAsyncLoad && bAsyncActorExists)
+		{
+			UpdateTestPassed(true);
+		}
+		else if (bOutputTestFailure)
+		{
+			UE_LOG(LogTemp, Warning, TEXT("Test invalid on this client - client correctly setup %d, async load on: %d, async actor exists: %d"), bAsyncTestCorrectlySetup, bAsyncLoad, bAsyncActorExists);
+			bOutputTestFailure = false;
```
jacquese, line 60: what is this meant to do? Won't this log once no matter what because it's initialised to true and we expect the async load to take longer than a tick?

USER_MANUAL.md
```diff
@@ -385,5 +385,14 @@ Manual steps:<br>
   1. Start the gym again and note that the level actors should be at their previous shutdown positions, not their original positions. You can do this repeatedly.
   1. The test has now passed. Don’t forget to revert the two settings changes you made before you run another test.
 
+##### Async Package Loading Gym
+Tests that async package loading works when activated.<br>
```
jacquese, line 389: This could do with an explanation of how it tests that- is it a client with async loading enabled connects and doesn't create an actor until the asset is loaded? Or actually I thought it was both a server and client thing.

jacquese (approved):
LGTM % couple of comments

Game/Source/GDKTestGyms/Private/AsyncPlayerController.cpp
m-samiec: This is meant to help debug if the test fails for any of the reasons above, but also prevent it spamming every frame. You're correct that the async actor isn't always available when this fires, so we get a false negative on the "correct" client. Will rethink and update.

USER_MANUAL.md
m-samiec: Yes, it tests that a client both doesn't have the class loaded in memory after begin play has been called, and manages to load an instance of that class at some later point in it's lifetime. This is definitely a thing on servers and clients, and this same logic could be run by a second non-auth server. I just picked a client because they're easier to spin up and attach to a deployment.

PR #311, opened 2021-02-10
Disable audio logs

joshuahuburn:
Audio logs were spamming our tests and making the results un-readable. This simply prevents that spam.
Example before: https://buildkite.com/improbable/unrealgdk-premerge/builds/22314
Example after: https://buildkite.com/improbable/unrealgdk-premerge/builds/22325


mironec:
🔊 🚫

mironec (approved):

PR #314, opened 2021-02-11
Override test settings

Vix2021:
A summary of the selected approach is:
- There can be a base optional config file to override settings for all maps e.g. `TestOverridesBase.ini`
- Each map can have an optional config file to override settings e.g. `TestOverrides<MapName>.ini`
- The current format is used, the settings class name and property name to set the value.
- Automatically override and revert the settings from a file using the save and load config functions.

For full details see: [Design Doc](https://docs.google.com/document/d/10HV70-j4dgeMPxnKWuPD_1PViBc24WmGj9YR7xqMhZ8/edit#)

The TestGym Changes required to support this are as follows:
- create an ini file for each map that requires to override settings
- remove the explicit setting of the number of clients in test -> PredictedGameplayCuesTest is in SpatialZoningMap so I created a new ini file for the map.
- update maps with new tests
    - EventTracingSettingsOverride (not in 4.25 ->done in 4.26)
    - SpatialSnapshotTestPart1SettingsOverride (done 4.25)
    - SpatialSnapshotTestPart2SettingsOverride (done 4.25)
    - SpatialTestNetReferenceSettingsOverride (done 4.25)
    - SpatialComponentSettingsOverride (not in 4.25 -> generated map -> added code)
    - SpatialAuthoritySettingsOverride (not in 4.25 -> generated map -> added code ->commented out as requires generated config files https://improbableio.atlassian.net/browse/UNR-5020)


Game/Config/TestOverridesControllerPossessPawnGym.ini
```diff
@@ -0,0 +1,6 @@
+[/Script/UnrealEd.LevelEditorPlaySettings]
```
mironec, line 1: By the way, @improbable-valentyn this will need coordination with @Vix2021 depending on which one goes in first, as you are moving the tests to use generated maps, and this will touch that.
I created `feature/gen-settings` as an experiment to generate these configs automatically using the commandlet. There is one small annoyance, which is that these configs are next to the normal ones, so show up in git and pollute things a bit. Not sure what we want to do with it and whether it's acceptable.
One of the alternatives could be to go at it from the opposite end, and make the test setting override PR check if the map is generated and override settings based on an in-memory string. Opinions welcome.

Vix2021: We can have a separate folder for the generated configs if we want a simple solution.
I would need to investigate if it is possible to override settings based on an in-memory string. I would prefer to do this as a separate task so these changes can be committed.

mironec: Created a draft PR with what I had in mind, just so people can look at the changes: https://github.com/spatialos/UnrealGDK/pull/2912

oblm:
I've confirmed that this functionality works by implementing it in my own test. The only defect that I noticed while using this feature if I change an `OverridesBase.ini` I then have to reboot the editor for that to take effect. Is that an known limitation and, if so, is it documented anywhere?

Once the above is address I'm happy to approve this and the accompanying PRs.

oblm:
> I've confirmed that this functionality works by implementing it in my own test. The only defect that I noticed while using this feature if I change an `OverridesBase.ini` I then have to reboot the editor for that to take effect. Is that an known limitation and, if so, is it documented anywhere?
> 
> Once the above is address I'm happy to approve this and the accompanying PRs.

@Vix2021 has now resolved this with an UnrealEngine fix.

oblm (approved):

Vix2021:
> I've confirmed that this functionality works by implementing it in my own test. The only defect that I noticed while using this feature if I change an `OverridesBase.ini` I then have to reboot the editor for that to take effect. Is that an known limitation and, if so, is it documented anywhere?
> 
> Once the above is address I'm happy to approve this and the accompanying PRs.

Thanks for highlighting this usability issue. I am requesting the config files are loaded for every test run but they are being cached inside the Engine in the `FConfigCacheIni:: Find` function. I have implemented a fix for this to force a reload if they are test setting override config files.

PR #315, opened 2021-02-15
Re-enable cross server ability activation test

KeyMaster-:
Also set the correct required client count on it while I'm here.


Game/Source/GDKTestGymsFunctionalTests/Tests/CrossServerAbilityActivationTest/CrossServerAbilityActivationTest.cpp
```diff
@@ -24,6 +24,7 @@ ACrossServerAbilityActivationTest::ACrossServerAbilityActivationTest()
 {
 	Author = TEXT("Tilman Schmidt");
 	Description = TEXT("Tests that gameplay abilities can be activated across server boundaries.");
+	SetNumRequiredClients(1);
```
mironec, line 27: This will be going away with Victoria's PR, but either way is fine, I think

mironec (approved):

Game/Source/GDKTestGymsFunctionalTests/Tests/CrossServerAbilityActivationTest/CrossServerAbilityActivationTest.cpp
KeyMaster-: Might be that we want to clean out this API and calls to it as a follow-up to that PR then.

PR #317, opened 2021-02-19
[UNR-4833] Fix gameplay cues test

KeyMaster-:
- Moved to single worker map.
- Changed cue tags to avoid cues in the game project triggering instead of the test cues.
- Fixed warning happening due to the ability not being ended (due to passing in the wrong actor info).


mironec:
The test ran and passed! Seems good to me.
(Don't know what this GAS stuff is)

mironec (approved):

Game/Source/GDKTestGymsFunctionalTests/Tests/PredictedGameplayCuesTest/GA_ApplyCueEffect.cpp
```diff
@@ -29,7 +29,7 @@ void UGA_ApplyCueEffect::ActivateAbility(const FGameplayAbilitySpecHandle Handle
 	{
 		if (!CommitAbility(Handle, ActorInfo, ActivationInfo))
 		{
-			EndAbility(CurrentSpecHandle, CurrentActorInfo, CurrentActivationInfo, false, true);
+			EndAbility(Handle, ActorInfo, ActivationInfo, false, true);
```
mironec, line 32: Ultra nit and I have no idea what I am looking at:
Is this function now only dependent on input args? Is it static?

KeyMaster-: Yes on only being dependent on input args, but no on the static bit. This is an overridden function from UGameplayAbility, so we can't change anything about the signature.

PR #323, opened 2021-03-04
Setup ET sampling settings for latency gym 

MatthewSandfordImprobable:
Lot's of testing commits in the PR. The final state of the Sampling settings it to set the default sampling rate to 0 and add overrides for all latency events using a probability of 1. This means only events concerned with latency traces will be logged.

Additionally I fixed an event in LatencyPlayerCharacter_BP.uasset that was using the wrong string for it's "type".

TestGymBuildKite PR: https://github.com/improbable/TestGymBuildkite/pull/235

New settings:
image

image
We used to use event type:"latency.cross_server_rpc". This was changed to: "latency.cross_server_property".


m-samiec:
Can you post a screen shot of what the blueprint changes look like?

MatthewSandfordImprobable:
> Can you post a screen shot of what the blueprint changes look like?

Added

m-samiec (approved):

joshuahuburn (approved):

PR #324, opened 2021-03-04
Remove maps for tests that have C++ generated ones

improbable-valentyn:
GDK PR: https://github.com/spatialos/UnrealGDK/pull/2940


simonsarginson (approved):

mironec (approved):
I didn't cross-reference whether the correct maps got removed, placing my trust in you

mironec:
Actually - aren't there some config files that we can now remove? Since they would be defined with the generated maps

improbable-valentyn:
@mironec good point, some of the configs can be cleaned up, I'll address them after this is merged though.

PR #326, opened 2021-03-09
UNR-4915 fix/limit-default-client-permission2

simonsarginson:
remove unused configs, should have been done in original branch fix/limit-default-client-permission


MatthewSandfordImprobable:
Nice! I would check around and see if we reference these in any documentation. Ollie might be a good person to ask. If we do reference in docs them we should probably add a ticket to update them to use more modern features.

MatthewSandfordImprobable (approved):

PR #331, opened 2021-03-17
Reduce latency et sampling

MatthewSandfordImprobable:
Sampling rate 0.25 -> 0.025. Through experimentation I found 0.25 way too much data to process. 0.025 still may prove too high but it is a stop gap for now.

Added missing end trace event for properties. This allows us to filter all traces for the start and end events and pull out a latency (time difference) between events.


improbable-danny:
Do we still get the same number as samples as stackdriver? I think it's around 12k of each trace path?

improbable-danny (approved):
Happy with the changes but think we should ensure we have the same number of samples per trace path as we previously had (for statistical purposes)

joshuahuburn (approved):

PR #332, opened 2021-03-19
[MBL-222]Call memreport on a configurable interval on the server worker

Jane-XuJingyi:
To profile memory, we can get memreport via Worker Flag.
The Worker Flag name is `mem_report`. Its value represents the interval(seconds) between two reports.  
The output files are saved as `worker_nanny\data\workers\UnrealWorker\GDKTestGyms\Saved\Profiling\MemReports\Map-Platform-ProfilesessionTimestamp\workerID-SnapshotTimestamp.memreport`.
Reports during the same profile session will be saved to the same folder.

Currently, you have to add this flag in json files to use this feature. Don't forget to set file restore directory `"--extra_node_paths_to_download=workerNode/worker_nanny/data/workers/UnrealWorker/GDKTestGyms/Saved/Profiling/"` in `...\TestGymBuildkite\ci\config\regisseur_flags\gdk.json`

Jira Ticket: [MBL-222](https://improbableio.atlassian.net/browse/MBL-222)
NFR test result: https://buildkite.com/improbable/product-research-benchmarks-run/builds/21834#59511ea7-67e1-4559-a23b-b22b47fd67dd


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -767,6 +795,12 @@ void ABenchmarkGymGameModeBase::SetStatTimer(const FString& TimeString)
 }
 #endif
 
+void ABenchmarkGymGameModeBase::SetMemReportIntervalTimer(const FString& TimeString)
+{
+	MemReportIntervalTimer.SetTimer(FCString::Atoi(*TimeString));
+	UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("MemReport Interval is set to %s seconds"), *TimeString);
```
jayprobable, line 801: This will add a log line every time SetMemReportIntervalTimer is called, and this get's called everytime the timer fires, it feels unnecessary. Recommend to add an int32 member for the interval and when processing the flag convert the string to int32 and log the value one time.

jayprobable (changes requested):
Looks like this code is following the SetStatTimer, but I think that code will need to change when you implement MBL-223, as it will want to repeated reset the timer.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -272,6 +276,23 @@ void ABenchmarkGymGameModeBase::Tick(float DeltaSeconds)
 		StatStopFileTimer.SetTimer(999999);
 	}
 #endif
+	if (MemReportIntervalTimer.HasTimerGoneOff())
+	{
+		if (MemReportInterval>0) //Avoid execute memreport each tick before the worker flag take effect
```
jeffxu-improbable, line 281: Why don't we check `MemReportInterval > 0` first, since it looks like a switch of this feature?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -272,6 +276,23 @@ void ABenchmarkGymGameModeBase::Tick(float DeltaSeconds)
...
+		{
+			FString Cmd = TEXT("memreport -full");
+			if (GetDefault<UGeneralProjectSettings>()->UsesSpatialNetworking())
```
jeffxu-improbable, line 284: Do we need support native unreal mode?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
Jane-XuJingyi: Modified the SetMemReportIntervalTimer to InitMemReportIntervalTimer. It will be called before Tick() when the flag is set and save the interval to an int32 member.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
Jane-XuJingyi: When SpatialOS is disabled, the command will follow the native Unreal naming rule as `Map-Platform-SnapshotTimestamp.memreport`. Both native and Spatial reports will be saved to the same file folder named with `Map-Platform-ProfileSessionTimestamp\`

jayprobable (approved):

joanxie878 (approved):

PR #338, opened 2021-03-29
Trigger the new CI pipeline without branch detection

joshuahuburn:
This will allow the new CI pipeline to pick up TestGyms builds.


.buildkite/premerge.steps.yaml
```diff
@@ -31,6 +31,42 @@ script_runner: &script_runner
...
 steps:
+  # New build pipeline
+  # Trigger a 4.26 build
+  - trigger: "unrealgdkbuild-ci"
+    label: "testgyms-ci-426 ${BUILDKITE_MESSAGE}"
+    async: true
+    build: 
+      env: 
+        BUILD_TYPE: "GDK"
+        GDK_BRANCH: "master" # TODO: Do we want auto-branch name detection?
```
mironec, line 42: of course

mironec: the way I would do it (probably not in the scope of the current ticket) is to allow special values in this field.
E.g. `{master}`, which the build pipeline would then interpret as "match branch name or default to master".
We can then have the branch matching logic localized in the CI repo and apply it to any repos we want.

joshuahuburn: Solid maybe right now

mironec (approved):

simonsarginson (approved):

PR #345, opened 2021-04-09
Disable Magic Leap plugins + Obsolete VR inputs plugins that were triggering WARNs

oblm:
* I made the plugin changes via the **Edit** > **Plugins** flow. I didn't manually edit the `.uproject` file.
  * These changes non-destructively disable Magic Leap.
  * They stop these errors triggering when you open the project or generate schema:
`Warning: MLSDK not found.  This likely means the MLSDK environment variable is not set.`
`Warning: VR disabled because ZI is not enabled.`
* I tested these before and after making the changes.
* I also deleted `+ActionMappings=(ActionName="ResetVR",Key=MotionController_Left_Grip1,bShift=False,bCtrl=False,bAlt=False,bCmd=False)`
  * Because it was triggering:
  * `LogInput: Warning: Action ResetVR uses invalid key MotionController_Left_Grip1.`
  * `LogInput: Warning: Use -RemoveInvalidKeys to remove instances of these keys from the action mapping.`

### DIFF
--- a/Game/Config/DefaultInput.ini
+++ b/Game/Config/DefaultInput.ini
@@ -22,4 +22,3 @@
 +AxisMappings=(AxisName="LookUp", Key=MouseY, Scale=-1.f)
 
 +ActionMappings=(ActionName="ResetVR", Key=R)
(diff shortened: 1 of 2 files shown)


mironec (approved):
Probably good, though I think removing the win64/mac supported platforms would be sufficient?
Maybe someone else has more context on the VR plugins.

PR #348, opened 2021-04-15
Fix labels and messages. Make new ci required. 

joshuahuburn:
Labels are used to identify which builds are required by GH. By generating unique labels for each branch we were spamming our GH UI with new builds to watch and also made it impossible to make the old-ci non-required without removing it. This fixes that.

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -35,10 +35,11 @@ steps:
   # New build pipeline
   # Trigger a 4.26 build
   - trigger: "unrealgdkbuild-ci"
-    label: "testgyms-ci-426 ${BUILDKITE_MESSAGE}"
-    async: true
+    label: "testgyms-ci-4.26"
+    async: false
     build:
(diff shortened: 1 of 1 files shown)


.buildkite/premerge.steps.yaml
```diff
@@ -35,10 +35,11 @@ steps:
   # New build pipeline
   # Trigger a 4.26 build
   - trigger: "unrealgdkbuild-ci"
-    label: "testgyms-ci-426 ${BUILDKITE_MESSAGE}"
-    async: true
+    label: "testgyms-ci-4.26"
```
mironec, line 38: are dots ok?

mironec (approved):

.buildkite/premerge.steps.yaml
joshuahuburn: Confirmed they are ok. Gives us consistency throughout the repos now too.

simonsarginson (approved):

PR #350, opened 2021-04-15
Remove old CI

joshuahuburn:
Realised there's no point of running old CI for TestGyms as all it does is call new CI for GDK.
The new implementation calls the UnrealGDKBuild repo as desired (new ci) and so no need to do it twice.

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -1,35 +1,26 @@
 ---
 ci_version: &ci_version "1.0"
+
 # This is designed to trap and retry failures because agent lost
 # connection. Agent exits with -1 in this case.
 agent_transients: &agent_transients
   exit_status: -1
   limit: 3
+
 # BK system error
 bk_system_error: &bk_system_error
   exit_status: 255
   limit: 3
+
 # job was interrupted by a signal (e.g. ctrl+c etc)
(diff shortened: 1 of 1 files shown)


.buildkite/premerge.steps.yaml
```diff
@@ -1,35 +1,26 @@
...
-    - "working_hours_time_zone=london"
-  retry:
-    automatic:
-      - <<: *agent_transients
-      - <<: *bk_system_error
-      - <<: *bk_interrupted_by_signal
+# Hook failure
+bk_hook_failure: &bk_hook_failure
+  exit_status: 127
+  limit: 3
```
mironec, line 23: we don't need this if we only trigger, no?

mironec (approved):
One comment, otherwise good

.buildkite/premerge.steps.yaml
joshuahuburn: I've actually seen the pre-checkout hooks fail in all repos, this auto-retries for that too.

simonsarginson (approved):

.buildkite/premerge.steps.yaml
mironec: But we don't run any steps here?

mironec: I guess only the premerge-trigger, but that wouldn't be affected by this anyway

PR #354, opened 2021-04-21
feature/unr-5296-extend-test-setting-overrides-for-groups-of-maps-425

Vix2021:
- Removed unnecessary config files and renamed two to use as new map config files.
    - many of these were removed because they are no longer needed as the tests are now generated
    - some became the new map config files
- Updated maps with group config filenames
   - most of these done on the v4.25 version of Unreal but the EventTracing maps were not accessible so had to be changed on v4.26 
image


m-samiec (approved):
Nice - looks much cleaner.

PR #375, opened 2021-05-12
Cleaning up Actor count metric and making it functional in zoning

MatthewSandfordImprobable:
### **Actor Count Modifications**

This PR aims to clean up some of the redundancies in our actor count logic on the game mode. It also aims to allow multi-worker tests to make use of the actor count metric.

The auth game mode server will now report the actor counts for all other workers and output totals as well:
image

We can now again add arbitrary actors to the actor count metric using ABenchmarkGymGameModeBase::AddExpectedActorCount.

Key Points:
- CounterComponent removed
- Game mode now counts auth actors.
- All servers send cross server RPCs to register their actor counts.

Have yet to test a zoning NFR with actor count metric enabled. Will do in the near future.

NFR RUN: https://buildkite.com/improbable/unrealgdk-nfr/builds/5079

### **Blueprint fix**

Addtionally, one fix in blueprint. We used to start spawning drop cubes in the BeginPlay of the SpawnDestroyCube component. If an actor crossed a boundary and the new auth server had already created the actor, we would not continue to spawn cubes. Whenever the authority changes on an NPC or player we ask the SpawnDestroyCube component to start spawning cubes. The 
newly non auth server will stop spawning cubes automatically as part of the existing blueprint logic.

Removed this from the SpawnDestroyCube_BP:
image

Start spawning process from the BenchmarkNPC_BP and PlayerCharacter_BP on authority gained.
image

TestGymBuildkite PR: https://github.com/improbable/TestGymBuildkite/pull/265


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -91,11 +91,8 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, ActorMigrationCheckDelay(5*60)
 	, PrintMetricsTimer(10)
 	, TestLifetimeTimer(0)
+	, TickActorCountTimer(0)
 	, bHasRequiredPlayersCheckFailed(false)
-	, SmoothedTotalAuthPlayers(-1.0f)
-	, SmoothedTotalAuthNPCs(-1.0f)
-	, SmoothedTotalAuthSimPlayers(-1.0f)
-	, RequiredPlayerReportTimer(10 * 60)
```
MatthewSandfordImprobable, line 98: This timer removed as we use Constants->ActorCheckDelay (which has the same delay)

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -315,6 +312,7 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
 
 	if (RequiredPlayerCheckTimer.HasTimerGoneOff() && !DeploymentValidTimer.HasTimerGoneOff())
 	{
+		int SmoothedTotalAuthPlayers = TotalActorCounts[SimulatedPawnClass].GetSmoothedTotal();
```
MatthewSandfordImprobable, line 315: We could consider getting rid of the TickPlayersConnectedCheck all together. Currently it simply checks the actor count of the player BP. Admittedly it is using the smoothed value rather than absolute... but I am sure we could come up with a way of configuring this in the actor count metric.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -463,20 +454,35 @@ void ABenchmarkGymGameModeBase::TickActorCountCheck(float DeltaSeconds)
 {
 	const UNFRConstants* Constants = UNFRConstants::Get(GetWorld());
 	check(Constants);
-	FMetricTimer TempActorCheckDelay = Constants->ActorCheckDelay;
 	// This test respects the initial delay timer in both native and GDK
-	if (TempActorCheckDelay.HasTimerGoneOff() &&
+	if (Constants->ActorCheckDelay.HasTimerGoneOff() &&
+		TickActorCountTimer.HasTimerGoneOff() &&
```
MatthewSandfordImprobable, line 459: TickActorCountTimer was added to allow us to set the frequency at which we wish to report actor count data. 
Preferable to re-using the Constants->ActorCheckDelay as these are "constant" timers (timers that should not really be modified). We could technically re-use it as the old code attempted to do...

Game/Source/GDKTestGyms/Public/CounterComponent.h
```diff
@@ -14,6 +14,12 @@ class GDKTESTGYMS_API UCounterComponent : public UActorComponent
 {
 	GENERATED_BODY()
 
+	struct FActorCountInfo
+	{
+		int32 TotalCount;
+		int32 AuthCount;
```
MatthewSandfordImprobable, line 20: Capturing both auth and total actors here. I don't think we actually require the total count anymore.

If we want to log actor counts on clients, we might want to re add the log that was in UCounterComponent::UpdateCachedActorCounts. Then total count would be useful. 

Interested to hear peoples opinions.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: `TickActorCountTimer` ?
Looks like that has a delay of 5 mins

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -461,22 +452,54 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
-		const int32 ActualCountSimulate = GetActorClassCount(ExpectedSimPlayersCount.ActorClass);
-		const int32 ActualCountNPCs = GetActorClassCount(ExpectedNPCsCount.ActorClass);
-		GenerateTotalNumsForActors(WorkerID, World, ExpectedSimPlayersCount, MapAuthoritativeSimPlayers, SmoothedTotalAuthSimPlayers, ActualCountSimulate, false);
-		ReportAuthoritativeNPCs(WorkerID, World, ActualCountNPCs);
-		TempActorCheckDelay.SetTimer(1);
+		if (WorkerID.IsEmpty())
+		{
+			return;
+		}
+		TMap<TSubclassOf<AActor>, FActorCountInfo>& ThisWorkerActorCounts = WorkerActorCounts.FindOrAdd(WorkerID);
```
m-samiec, line 466: `TMap<TSubclassOf<AActor>, FActorCountInfo>` is used a bit - worth creating a `using` statement for the type

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -461,22 +452,54 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
+
+				if (bActorCountInfoMissing || ActorCountInfo->GetTotal() != ActorCount)
+				{
+					ReportAuthoritativeActorCount(WorkerID, ActorClass, ActorCount);
+					ActorCountInfo->SetTotal(ActorCount);
+				}
+			}
+		}
+
+		const int32 TickActorCountFrequency = 1; /*seconds*/
```
m-samiec, line 496: Is it seconds or frequency? :)

Game/Source/GDKTestGyms/Public/CounterComponent.h
m-samiec: Seems useful to keep Total count around - valid for comparing against interest border size and subsequent performance impact

Game/Source/GDKTestGyms/Public/CounterComponent.h
```diff
@@ -23,7 +29,8 @@ class GDKTESTGYMS_API UCounterComponent : public UActorComponent
 	UPROPERTY(EditAnywhere, BlueprintReadWrite)
 	TArray<TSubclassOf<AActor>> ClassesToCount;
 
-	int32 GetActorClassCount(TSubclassOf<AActor> ActorClass) const;
+	int32 GetActorTotalCount(TSubclassOf<AActor> ActorClass) const;
```
m-samiec, line 32:
```suggestion
	int32 GetActorTotalCount(const TSubclassOf<AActor>& ActorClass) const;
```

m-samiec: here and elsewhere

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -823,50 +812,86 @@ int32 ABenchmarkGymGameModeBase::GetPlayerControllerCount() const
...
+	}
+
+	if (bLogActorCountDetails && TempTotalActorCounts.Num() > 0)
+	{
+		UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("--- Actor Count Totals ---"));
+	}
+
+	for (const auto& ActorCountPair : TempTotalActorCounts)
+	{
+		const TSubclassOf<AActor> ActorClass = ActorCountPair.Key;
```
m-samiec, line 871:
```suggestion
		const TSubclassOf<AActor>& ActorClass = ActorCountPair.Key;
```

m-samiec (commented):
Changes look good - can we get a full NFR run with this branch?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Yes, a slight delay before sending ReportAuthoritativeActorCount RPCs was necessary.  Admittedly 5 minutes is probably a lot larger than needed. We only send ReportAuthoritativeActorCount if locally the actor count changes. For "stable" actors where their count does not change, only a single RPC is sent as the actor count does not change locally. I found that without a delay at the start of the deployment, the auth server was not reporting the correct counts. I do not have a explanation for this but I put it down to setting up the deployment (servers starting, game mode authority being delegated etc). I reckon the delay could be like 10 seconds and it would be fine. As long as the delay is less then 10 mins the tests should function the same. We might want earlier logging early in the deployment though (before the actual testing starts). We can do more investigation into why the delay is needed if you like. I will reduce the delay to a minute though to give us a bit more logging.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -461,22 +456,55 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
+				const int32 ActorCount = GetActorAuthCount(ActorClass);
+
+				FActorCountInfo* ActorCountInfo = ThisWorkerActorCounts.Find(ActorClass);
+				bool bActorCountInfoMissing = ActorCountInfo == nullptr;
+				if (bActorCountInfoMissing)
+				{
+					ActorCountInfo = &ThisWorkerActorCounts.Add(ActorClass);
+				}
+
+				if (bActorCountInfoMissing || ActorCountInfo->GetTotal() != ActorCount)
```
MatthewSandfordImprobable, line 493: Check bActorCountInfoMissing to report if this is our first time.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -179,7 +179,7 @@ void ABenchmarkGymGameMode::BuildExpectedActorCounts()
 
 	const int32 TotalDropCubes = TotalNPCs + ExpectedPlayers;
 	const int32 DropCubeCountVariance = FMath::CeilToInt(TotalDropCubes * 0.1f) + 2;
-	AddExpectedActorCount(ExpectedDropCubeCount, DropCubeClass, TotalDropCubes, DropCubeCountVariance);
+	AddExpectedActorCount(DropCubeClass, TotalDropCubes - DropCubeCountVariance, TotalDropCubes + DropCubeCountVariance);
```
improbable-danny, line 182: Should this be +- DropCubeCountVariance*0.5?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -14,34 +14,68 @@ UCLASS()
...
+
+		void SetTotal(int32 InTotal)
+		{
+			Total = InTotal;
+			SmoothedTotal = SmoothedTotal * 0.9f + Total * 0.1f;
+		}
 
-	struct FExpectedActorCount
+		int32 GetTotal() const { return Total; }
+		int32 GetSmoothedTotal() const { return FMath::RoundToInt(SmoothedTotal); }
```
improbable-danny, line 39: Nit: Int truncation here might neglect 0.5 around the min and allow 0.5 more around max 
```suggestion
		int32 GetSmoothedTotal() const { return FMath::RoundToInt(SmoothedTotal + 0.5f); }
```

improbable-danny (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -147,14 +144,15 @@ void ABenchmarkGymGameModeBase::TryInitialiseExpectedActorCounts()
 
 void ABenchmarkGymGameModeBase::BuildExpectedActorCounts()
 {
-	AddExpectedActorCount(ExpectedNPCsCount, NPCClass, TotalNPCs, 1);
-	AddExpectedActorCount(ExpectedSimPlayersCount, SimulatedPawnClass, ExpectedPlayers, 1);
+	// Zoning scenarios can report actor count numbers slightly higher than the expected number so add a little slack.
```
simonsarginson, line 147: Why?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
simonsarginson: Its a rounding function it does not truncate? 
`	static FORCEINLINE int32 RoundToInt(float F)
	{
		return FloorToInt(F + 0.5f);
	}`

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Since the server report their actor counts at different times, when there is a transition of actors between zones, there will be some sync issues. In one case, the server gaining authority over the actor may report it's actor count before the server losing authority. In this case the calculated actor count would be high than expected. Will add a comment to summarize this.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
MatthewSandfordImprobable: FMath::RoundToInt will round 0.5 up. Perhaps I miss understand?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: Yes and no. I think the actual variance is good here. Perhaps DropCubeCountVariance  should be rename for clarity though. Maybe just HalfDropCubeCountVariance? What do you think?

Game/Source/GDKTestGyms/Private/CounterComponent.cpp
```diff
@@ -21,32 +21,53 @@ void UCounterComponent::TickComponent(float DeltaTime, ELevelTick TickType, FAct
 	if (Timer <= 0.0f)
 	{
 		Timer = ReportFrequency;
-		UpdateCachedActorCounts();
+		UpdateCachedAuthActorCounts();
 	}
 }
 
-int32 UCounterComponent::GetActorClassCount(TSubclassOf<AActor> ActorClass) const
+int32 UCounterComponent::GetActorTotalCount(const TSubclassOf<AActor>& ActorClass) const
```
joanxie878, line 28: Seems like this function is not used anywhere

joanxie878 (approved):

Game/Source/GDKTestGyms/Private/CounterComponent.cpp
MatthewSandfordImprobable: This is true! The counter component now counts the number of actors a worker has interest in as well as the number of actors it has auth over. Currently we don't actually use the interest actor count anywhere...

I left this in as it seems like it could potentially be useful and is very lightweight. Mike also liked the idea of keeping this code around.

You are right though this could be removed... I am in two minds about it...

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -150,14 +148,16 @@ void ABenchmarkGymGameModeBase::TryInitialiseExpectedActorCounts()
...
```
MatthewSandfordImprobable, line 151: Might not be true anymore

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
improbable-danny: Sounds fine, think I just misunderstood the previous usage!

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
improbable-danny: Does this comment need updating?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -331,17 +331,21 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
...
-			bHasRequiredPlayersCheckFailed = true;
-			// This log is used by the NFR pipeline to indicate if a client failed to connect
-			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Client connection dropped. Required %d, got %.1f"), *NFRFailureString, RequiredPlayers, SmoothedTotalAuthPlayers);
-		}
-		else
-		{
-			RequiredPlayerCheckTimer.SetTimer(10);
-			// Useful for NFR log inspection
-			NFR_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("All clients successfully connected. Required %d, got %.1f"), RequiredPlayers, SmoothedTotalAuthPlayers);
+			if ((*ActorCount) < RequiredPlayers)
```
improbable-danny, line 337: Nit + opinion: Don't feel brackets are nessesary
```suggestion
			if (*ActorCount < RequiredPlayers)
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -681,57 +675,91 @@ void ABenchmarkGymGameModeBase::OnRepTotalNPCs()
...
-		MapAuthoritativePlayers.Emplace(WorkerID, AuthoritativePlayers);
-		int32 TotalPlayers = 0;
-		for (const auto& KeyValue : MapAuthoritativePlayers)
+		if (Actor->HasAuthority())
 		{
-			TotalPlayers += KeyValue.Value;
+			AuthCount++;
 		}
-		if (MapAuthoritativePlayers.Num() == NumWorkers)
+		else if (PackageMap != nullptr)
```
improbable-danny, line 736: Should we `ensure` this?

MatthewSandfordImprobable: Nah, but we defo need to make it work in native :/

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -331,17 +331,21 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
 
 	if (RequiredPlayerCheckTimer.HasTimerGoneOff() && !DeploymentValidTimer.HasTimerGoneOff())
 	{
-		if (SmoothedTotalAuthPlayers < RequiredPlayers)
+		const int32* ActorCount = TotalActorCounts.Find(SimulatedPawnClass);
+		if (ActorCount != nullptr)
```
m-samiec, line 335: If it doesn't find it, seems like that should be a failure also. Can also do the declaration within the if statement.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -486,22 +483,19 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
-		GenerateTotalNumsForActors(WorkerID, World, ExpectedSimPlayersCount, MapAuthoritativeSimPlayers, SmoothedTotalAuthSimPlayers, ActualCountSimulate, false);
-		ReportAuthoritativeNPCs(WorkerID, World, ActualCountNPCs);
-		TempActorCheckDelay.SetTimer(1);
+
+		if (HasAuthority())
+		{
+			ActorCountReportId++;
+			UpdateAndReportActorCounts();
+
+			const int32 TickActorCountPeriod = 10; /*seconds*/
```
m-samiec, line 496: Hah, I'd just put it in the variable name :)

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -681,57 +675,91 @@ void ABenchmarkGymGameModeBase::OnRepTotalNPCs()
...
-	return CounterComponent->GetActorClassCount(ActorClass);
+	UpdateAndReportActorCounts();
 }
 
-void ABenchmarkGymGameModeBase::SetLifetime(int32 Lifetime)
+void ABenchmarkGymGameModeBase::UpdateAndReportActorCounts()
 {
-	if (TestLifetimeTimer.SetTimer(Lifetime))
+	const UWorld* World = GetWorld();
+	const FString WorkerID = GetGameInstance()->GetSpatialWorkerId();
```
m-samiec, line 686: Won't this fail on native runs now?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Fair LOL

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -852,78 +880,91 @@ void ABenchmarkGymGameModeBase::InitMemReportTimer(const FString& MemReportInter
...
-					++Count;
-				}
+				bAllWorkersInSync = false;
+				break;
 			}
 		}
+
+
+		ActorCountReportedIds.Empty();
+		if (bAllWorkersInSync)
```
m-samiec, line 906: Little concerned that if workers get desynced in some way, this will never fire, and we'll get a potential false positive. Thoughts?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: They still have the potential to go over but from what I have seen the numbers vary a lot less (I have only ever seen the numbers go over by 1)

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: As in, if we are always desynced, will never check the actors count is invalid so the actor count metric wont fail so the test would pass. 

I think we do need to ensure the workers are in sync, otherwise there is no point looking at the reported values.
It's worth pointing out if there is a desync, subsequent actor counts should not be affected unless there are continuous desyncs or very frequent desyncs.  

We could fail the test though if we see desyncs over a threshold. It would be nice to actually see how often this case occurs.

m-samiec: Yeah I think we should just add a timer that's perhaps 2x the actor sync reporting time, and each time the auth server processes `UpdateAndCheckTotalActorCounts` we reset the timer. If it ever times out, it can fail the actor count metric but with a custom reason why.

MatthewSandfordImprobable: Custom reason being a custom log or a new metric/ expectation?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -681,57 +672,90 @@ void ABenchmarkGymGameModeBase::OnRepTotalNPCs()
...
 {
-	if (TestLifetimeTimer.SetTimer(Lifetime))
+	const UWorld* World = GetWorld();
+	bool bSpatialEnabled = USpatialStatics::IsSpatialNetworkingEnabled();
+	const FString WorkerID = bSpatialEnabled ? GetGameInstance()->GetSpatialWorkerId() : TEXT("Worker1");
+	if (WorkerID.IsEmpty())
 	{
-		TestLifetimeTimer.SetLock(true);
-		UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("Setting NFR test lifetime %d"), Lifetime);
+		return;
```
m-samiec, line 687: This seems like a fatal also

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -852,78 +876,91 @@ void ABenchmarkGymGameModeBase::InitMemReportTimer(const FString& MemReportInter
 }
 #endif
 
-int32 ABenchmarkGymGameModeBase::GetPlayerControllerCount() const
+void ABenchmarkGymGameModeBase::ReportAuthoritativeActorCount_Implementation(const int32 ActorCountId, const FString& WorkerID, const TArray<FActorCount>& ActorCounts)
```
m-samiec, line 879: nit - could we change ActorCountId to ActorCountReportIdx here and elsewhere?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -852,78 +876,91 @@ void ABenchmarkGymGameModeBase::InitMemReportTimer(const FString& MemReportInter
...
-				{
-					++Count;
-				}
+				bAllWorkersInSync = false;
+				break;
 			}
 		}
+
+
+		ActorCountReportedIds.Empty();
```
m-samiec, line 901: This seems wrong doesn't it? Shouldn't it be done only if `bAllWorkersInSync` is true? Also perhaps it isn't necessary, as you'd write over the top of previous values anyway.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -852,78 +876,91 @@ void ABenchmarkGymGameModeBase::InitMemReportTimer(const FString& MemReportInter
...
+		const FString& WorkerId = WorkerPair.Key;
+		const ActorCountMap& SpecificWorkerActorCounts = WorkerPair.Value;
+
+		UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("--- Actor Count for Worker: %s ---"), *WorkerId);
+
+		for (const auto& ActorCountPair : SpecificWorkerActorCounts)
 		{
-			CurNums += KeyValue.Value;
+			const TSubclassOf<AActor>& ActorClass = ActorCountPair.Key;
+			const int32& ActorCount= ActorCountPair.Value;
```
m-samiec, line 931:
```suggestion
			const int32& ActorCount = ActorCountPair.Value;
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Only reason this was here was that I found the GetGameInstance()->GetSpatialWorkerId() was not set early in the deployment.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -331,17 +332,18 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
...
+			NFR_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("All clients successfully connected. Required %d, got %d"), RequiredPlayers, (*ActorCount));
 		}
 		else
 		{
-			RequiredPlayerCheckTimer.SetTimer(10);
-			// Useful for NFR log inspection
-			NFR_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("All clients successfully connected. Required %d, got %.1f"), RequiredPlayers, SmoothedTotalAuthPlayers);
+			bHasRequiredPlayersCheckFailed = true;
+			// This log is used by the NFR pipeline to indicate if a client failed to connect
+			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Client connection dropped. Required %d, got %d"), *NFRFailureString, RequiredPlayers, (*ActorCount));
```
improbable-danny, line 346: `ActorCount` can be nullptr here

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
improbable-danny: Is it worth logging just in case?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -486,22 +481,27 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
-		const FString WorkerID = GetGameInstance()->GetSpatialWorkerId();
-		const int32 ActualCountSimulate = GetActorClassCount(ExpectedSimPlayersCount.ActorClass);
-		const int32 ActualCountNPCs = GetActorClassCount(ExpectedNPCsCount.ActorClass);
-		GenerateTotalNumsForActors(WorkerID, World, ExpectedSimPlayersCount, MapAuthoritativeSimPlayers, SmoothedTotalAuthSimPlayers, ActualCountSimulate, false);
-		ReportAuthoritativeNPCs(WorkerID, World, ActualCountNPCs);
-		TempActorCheckDelay.SetTimer(1);
+
+		if (HasAuthority())
+		{
+			ActorCountReportIdx++;
```
improbable-danny, line 491: I wonder if clearing the count map here would make it easier to track count updates not coming back (stress scenarios?)

improbable-danny: or setting them to 0, and letting that fail them in 10 seconds time

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Sure.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -333,17 +334,23 @@ void ABenchmarkGymGameModeBase::TickPlayersConnectedCheck(float DeltaSeconds)
...
 	{
-		if (SmoothedTotalAuthPlayers < RequiredPlayers)
+		const int32* ActorCount = TotalActorCounts.Find(SimulatedPawnClass);
+
+		if (ActorCount == nullptr)
 		{
-			bHasRequiredPlayersCheckFailed = true;
-			// This log is used by the NFR pipeline to indicate if a client failed to connect
-			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Client connection dropped. Required %d, got %.1f"), *NFRFailureString, RequiredPlayers, SmoothedTotalAuthPlayers);
+			NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Could not get Simulated Player actor count."), *NFRFailureString);
```
m-samiec, line 341: Add `bHasRequiredPlayersCheckFailed = true;`

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -488,22 +488,27 @@ void ABenchmarkGymGameModeBase::TickUXMetricCheck(float DeltaSeconds)
...
+
+		if (HasAuthority())
+		{
+			ActorCountReportIdx++;
+			UpdateAndReportActorCounts();
+
+			const int32 TickActorCountPeriodInSeconds = 10;
+			TickActorCountTimer.SetTimer(TickActorCountPeriodInSeconds);
+
+			TimeSinceLastCheckedTotalActorCounts += DeltaSeconds;
```
m-samiec, line 504: This will only increment on the frames that `TickActorCountTimer.HasTimerGoneOff()` is true, which is only every 10 seconds right? So can't use DeltaSeconds here, but need to use absolute game time. I'd recommend just switching this to a timer on the timermanager, that you reset in `ReportAuthoritativeActorCount`. Here's an example of how we used to use this same pattern in the heartbeat manager - https://github.com/spatialos/UnrealGDK/blob/925bba1ab2ab4ba1ed33758624478910ef5e2ecf/SpatialGDK/Source/SpatialGDK/Private/EngineClasses/SpatialNetConnection.cpp#L148

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -684,57 +689,94 @@ void ABenchmarkGymGameModeBase::OnRepTotalNPCs()
...
+			continue;
+		}
+
+		const FExpectedActorCountConfig& Config = Pair.Value;
+		if (Config.MinCount > 0)
+		{
+			int32 TotalCount = 0;
+			int32& AuthCount = ThisWorkerActorCounts.FindOrAdd(ActorClass);
+			GetActorCount(ActorClass, TotalCount, AuthCount);
+			NFR_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("Actor count for %s: ActorClass: %s Count: %d, AuthCount: %d"), *WorkerID , *ActorClass->GetName(), TotalCount, AuthCount);
```
m-samiec, line 723: WorkerID is redundant here because it'll be included in that workers logs

MatthewSandfordImprobable: Sure, although I did find it useful on the auth server, as there are a lot of actor counts being printed.

m-samiec: I'm not sure I follow - doesn't this only log the local worker's actor counts?

MatthewSandfordImprobable: Yes, but on the auth server there are logs for the total actor count as well as the auth actor counts for all server worker. Just thought it made sense to be explicit there to avoid any potential confusing.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -138,27 +137,89 @@ void ABenchmarkGymGameModeBase::BeginPlay()
...
-	Actor = FExpectedActorCount(ActorClass, ExpectedCount, Variance);
+	if (HasAuthority())
+	{
+		ActorCountReportIdx++;
+		UpdateAndReportActorCounts();
+
+		FTimerManager& TimerManager = GetWorld()->GetTimerManager();
+		if (!TimerManager.IsTimerActive(FailActorCountTimeoutTimerHandle))
+		{
+			float FailActorCountTimeout = 2.5f * UpdateActorCountCheckPeriodInSeconds;
```
m-samiec, line 195:
```suggestion
			const float FailActorCountTimeout = 2.5f * UpdateActorCountCheckPeriodInSeconds;
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -153,21 +171,26 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
...
+
 	void TickPlayersConnectedCheck(float DeltaSeconds);
 	void TickPlayersMovementCheck(float DeltaSeconds);
 	void TickServerFPSCheck(float DeltaSeconds);
 	void TickClientFPSCheck(float DeltaSeconds);
 	void TickUXMetricCheck(float DeltaSeconds);
-	void TickActorCountCheck(float DeltaSeconds);
 	void TickActorMigration(float DeltaSeconds);
 
+
```
m-samiec, line 192:
```suggestion
```

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.h
```diff
@@ -153,21 +171,26 @@ class GDKTESTGYMS_API ABenchmarkGymGameModeBase : public AGameModeBase
...
+
```
m-samiec, line 193:
```suggestion
```

m-samiec (approved):
Nice!

PR #377, opened 2021-05-13
UNR-5299 Destroy drop cube when player disconnects

jayprobable:
UNR-5299-bp1

UNR-5299-bp2

In the Uptime scenario we found that drop cube counts were increasing. After reviewing logs it became clear that when the SpawnDestroyCube component was destroyed (when a player disconnected) the existing spawned cube was not destroyed. Because the uptime scenario disconnects and reconnects SimPlayers we found over time the number of drop cubes continued to increase. Ultimately this results in thousands of drop cubes which ultimately overwhelm the uptime test.

This fix uses the End Play event of the SpawnDestroyCube component to issue a Cross Server RPC call to set the should self destruct flag of the drop cube.

Thanks to @jeffxu-improbable for recreating this locally to verify the behavior (before and after fix)


jeffxu-improbable (approved):
Verified on my local environment, it works!

joanxie878 (approved):

MatthewSandfordImprobable (approved):

PR #379, opened 2021-05-14
Update USER_MANUAL.md

oblm:
* Uncommenting `ReplicationDriverClassName` activates the Replication Graph.
* The Replication Graph's `GridSpatialization2D` node contains `ActorListFrequencyBuckets`, which by default have 3 buckets of actors, and sends updates to only one bucket per tick.
* This results in the cubes in the gym being updates every third tick, which causes them to move discontinuously (visually stuttering).
* This setting can be tweaked, but it is not in this gym.
* This note simply explains to users that they should not worry about the discontinuous motion of the cubes.


improbable-dmitrii (approved):
Thank you for updating this!

PR #383, opened 2021-05-17
[UNR-5548] Add Visual Logger manual test gym

martin-improbable:
Added a test gym (and supporting actor), to allow manual testing for actor visual logging in multi worker scenarios.

Added test notes to USER_MANUAL.


oblm (approved):
Thanks for the detailed steps.

PR #385, opened 2021-05-18
Bugfix/unr-5463

TeriDrummond:
https://improbableio.atlassian.net/browse/UNR-5463

Fixes bug with cubes not moving on first run of MultiworkerWorldComposition gym (or never working on a non-single-process run). 

The bug was that the cubes existed in sublevels which hadn't streamed in until after the Rep Graph had already iterated through all existing classes. The bug also happened in Unreal Native. 

The fix is to re-parent GymCube to ReplicatedActor (instead of Actor), because there is already code in the Rep Graph to handle classes that inherit from ReplicatedActor regardless of whether they've streamed in yet. 

This PR also has a change to the MultiWorkerSettings used for this test level to fix a separate problem, which is explained in detail in this PR: https://github.com/spatialos/UnrealGDK/pull/3129
I've also updated a handful of other SpatialZoning test levels which used these c++ Test WorkerSettings, to swap out the MultiworkerSettings they were using with new equivalents that are fully BPs, so that they can also launch in non-single-process.


m-samiec:
Was `MultiworkerWorldComposition.umap ` the only map that used a now-hidden test class?

TeriDrummond:
> 
> 
> Was `MultiworkerWorldComposition.umap ` the only map that used a now-hidden test class?

Good shout, there are a few maps that use those classes. Talked with Miron about what to do with them, we decided:
- These are pretty old and probably not used
- Intrepid might still want them as helpful examples
- I'll change them to use a BP equivalent of the worker settings, and try to make sure they still work by running them as tests in the session frontend

m-samiec (approved):
Could you update the PR description clarifying that you changed the other maps for the same reason?

PR #400, opened 2021-05-27
Bugfix/unr-5543

TeriDrummond:
Fix for https://improbableio.atlassian.net/browse/UNR-5543 -- a bug where the NPCs in the benchmark gym would stop being able to move after they'd crossed the zoning boundary. The problem was that the Blackboard values that their AI tree used to know which locations to run to were not set up to replicate across server workers, so as soon as they hit the boundary they'd get stuck with their default setting (which was "go to your current location"). 

The fix includes
- Set the "move to position" values to Handover
- When a controller is newly created (upon crossing the border), set the handed over values into its Blackboard
- Set the AIControllers to destroy when they lose authority (previously, a new one got created every time we crossed a border, so a ton would get created over time without being destroyed). This required reparenting the NPC Character blueprint because Unreal doesn't let you call Destroy on a Controller from within a BP, only within C++

**Note that the benchmark gym still has a problem with NPCs running across the border, because there is nothing in place to save the AI tree state across the border, so they end up looking janky. [NWX has a solution for AI migration ](https://github.com/improbable/Unreal-AI-migration-work)but if we want to apply that, we should file a new ticket to plan for it.


improbable-danny (approved):
Looks good, would be awesome if we could quick fix the issue of AI state handover for NFRs

MatthewSandfordImprobable:
Looks good! From blueprint, how do we know what to call ApplyBlackboardValues or do we just do it every time the controller is created?

TeriDrummond:
> 
> 
> Looks good! From blueprint, how do we know what to call ApplyBlackboardValues or do we just do it every time the controller is created?

Yeah! I'm doing it from BP right after the controller is created on the Character. 
image

TeriDrummond:
> 
> 
> Looks good, would be awesome if we could quick fix the issue of AI state handover for NFRs

Yeah, I'm going to try to get this working today. I have a couple ideas of how to make them sensibly continue moving after crossing the border so hopefully one of those ideas will work :P

MatthewSandfordImprobable (approved):

Game/Source/GDKTestGyms/Private/BenchmarkNPCCharacter.cpp
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Improbable Worlds Ltd, All Rights Reserved
+
```
joshuahuburn, line 2: Minor, double new lines in this file at start, after includes and end seem strange. Surprised our linter didn't pick this up

Game/Source/GDKTestGyms/Private/BenchmarkNPCCharacter.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include "BenchmarkNPCCharacter.h"
+
+
+void ABenchmarkNPCCharacter::OnAuthorityLost()
+{
+	Super::OnAuthorityLost();
+
+	if (Controller)
+	{
+		Controller->Destroy();
```
joshuahuburn, line 13: I'm wondering if this is something we should be doing for all AI Characters?

MatthewSandfordImprobable: Do you mean at some spatial level or project level?

I think you could potentially cache the controller on the actor as a non replicated property. If the NPC comes back to your zone you could reuse the old controller instead of creating a new one. If you loose interest in the NPC though, maybe we should always destroy the controller? I guess you could still cache the NPC entity Id and the controller and match the two back up when it comes back into view...

TeriDrummond: Yeah, I wasn't sure how much of this would be a game-specific implementation, though.

Game/Source/GDKTestGyms/Private/BenchmarkNPCCharacter.cpp
TeriDrummond: If that's something I need to set up locally it's likely I haven't done it yet :P This is just what the file had when Unreal created it for me.

Game/enc_temp_folder/7ce345a7f675f6fd8b5ac5ecee8c345/BlackboardValues.h
```diff
@@ -0,0 +1,24 @@
...
+struct GDKTESTGYMS_API FBlackboardValues 
+{
+	GENERATED_BODY()
+
+	UPROPERTY()
+	FVector TargetAValue; // Points to run betweeen
+	UPROPERTY()
+	FVector TargetBValue;
+	UPROPERTY()
+	bool TargetStateIsAOrB; // Keep track of current goal target, to continue towards it after handover. True for A, false for B
```
improbable-danny, line 19: nit: aligns more with comment :D 
```suggestion
	bool TargetStateIsA; // Keep track of current goal target, to continue towards it after handover. True for A, false for B
```

improbable-danny (approved):

joshuahuburn (approved):

PR #407, opened 2021-06-02
Fix generate schema error by replace handover with server only tag.

jeffxu-improbable:
Fix generate schema error.
Replace `Handover` with `Cond_ServerOnly` tag.

There was a time gap between `UNR-5447` and `UNR-5543`, so they missed it.
When raise PR, there was a pre merge check, which did not include another PR which was in review.


mironec (approved):
Thanks for doing this! Sorry that I missed it, realized after I merged.

Should be noted that the CI passed because this was intentionally left as a warning, rather than an error.

PR #408, opened 2021-06-02
[UNR-4146] Rename collision handover gym and change its description.

razvan-at-improbable:
BEFORE: (Wording is confusing on actual behavior, cube DIDN'T fall through the floor before)
image
AFTER:
image


ImprobableNic:
It looks like this could be automated.

razvan-at-improbable:
I made a ticket for improving it. I will include automation in the ticket.

PR #420, opened 2021-06-11
Generic grid spawning in test gyms project

MatthewSandfordImprobable:
GDK PR: https://github.com/spatialos/UnrealGDK/pull/3183
TestGymBuildkite PR: https://github.com/improbable/TestGymBuildkite/pull/275

### **Spawn Logic Changes:**

The aim is to allow the use of any configuration of grid based load balancing system in conjunction with the benchmark gym. NPC and player spawning logic has been adapted to accommodate any grid layout. 

Full NFR Run: https://buildkite.com/improbable/unrealgdk-nfr/builds/5694

PercentageSpawnPointsOnWorkerBoundaries can be varied to spawn more or less characters in zones or on boundaries. 
 
PercentageSpawnPointsOnWorkerBoundaries  = 0:
image

PercentageSpawnPointsOnWorkerBoundaries  = 0.5:
image

Setup that has multiple clusters on boundaries: 
image

### **Additional Changes:**

**AUptimeGameMode Changes**

I have changed AUptimeGameMode to inherit from ABenchmarkGymGameMode as there is a lot of overlap. It now does not need to read a number of worker flags/command line args as we read these values from the LB strategy directly in ABenchmarkGymGameMode. AUptimeGameMode extends ABenchmarkGymGameMode by adding an additional step into StartCustomNPCSpawning to spawn "beacons". AUptimeGameMode also ensures no clusters are on zoning boundaries by setting PercentageSpawnPointsOnWorkerBoundaries = 0.

**Fix Drop Cubes not Being Destroyed in Uptime Gym**
- Reimplemented a fix by Jay. We need to destroy the drop cube on EndPlay so that disconnected players do not leave drop cubes lying aroud.
image
image

**Spawn Drop Cubes in Uptime Gym**
- We recently changed the logic that started the spawning of drop cubes. This also needs to be done for the uptime character.
image

**Updated Game Modes to Specify Drop Cube Class**
- Update Uptime, Latency and Benchmark gyms blueprint game modes to specify the drop cube class rather than doing it from code.
image

**Updated Game Modes to Specify Drop Cube Class**
- Add assets for 2x2 zoning.
image
image

**Increased the Size of the 3x3 Zoning Config**
- The 3x3 zoning config had a world size not large enough to handle the required NPCs and Simplayer neatly. The world size was increased to compensate. This should have no affect on the load of the scenario.
image

**Ensuring Target points are initialized before Characters start moving **
- There is an existing issue where characters will initially move towards the center of the map. This is due to their behaviors trees being run without their target "walk between" points being setup. A blackboard value is now set to signify when it is appropriate for the behavior tree to start telling the character to move.
image

**Fixed calling cross server RPCs before NPCSpawn is ready**
-  When attempting to send a cross server rpc to spawn NPCs, it was possible for the RPC to get dropped due to the actor not "being ready". This could cause the total number of NPCs to be lower than needed. We will now retry when the spawner is not ready. This probably is not currently affecting us in the NFRs but only by chance.


Game/Source/GDKTestGyms/BenchmarkGymGameMode.h
```diff
@@ -12,6 +12,88 @@ DECLARE_LOG_CATEGORY_EXTERN(LogBenchmarkGymGameMode, Log, All);
 
 typedef TPair<TWeakObjectPtr<AController>, int> ControllerIntegerPair;
 
+UCLASS()
```
simonsarginson, line 15: I suggest adding a comment why you'd want to use this class?

simonsarginson:
is the cpp file missing?

MatthewSandfordImprobable:
> is the cpp file missing?

Logic should all be in BenchmarkGymGameMode.cpp.

simonsarginson:
> Logic should all be in BenchmarkGymGameMode.cpp.
Github's auto hide of big diffs got me : /

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -41,6 +351,29 @@ ABenchmarkGymGameMode::ABenchmarkGymGameMode()
...
+void ABenchmarkGymGameMode::BeginPlay()
+{
+	Super::BeginPlay();
+	SpawnManager = NewObject<USpawnManager>(this);
+
+	if (HasAuthority())
+	{
+		// It's possible for CheckCmdLineParameters may not get called on the game mode auth server if no players are spawned on the game mode auth server.
+		// In case this happens, ensure we still call CheckCmdLineParameters after a delay. The delay ensure worker flags are set.
+		FTimerManager& TimerManager = GetWorld()->GetTimerManager();
```
jeffxu-improbable, line 365: Perhaps we can just call `CheckCmdParameters` here instead?
`GameMode` will spawn NPC when it is ready.
There is no need to wait for players.

MatthewSandfordImprobable: So the reason I believe this needs to be on timer is so that we give a bit of time for the worker to receive worker flag values.
 CheckCmdParameters will start the logic to create PlayerStarts. To run this logic we need to know the PlayerDensity and ExpectedPlayer count so that we ensure we create clusters with the correct number of players.

I will test locally to see if BeginPlay is late enough so that we have received the worker flags. I remember this was a problem at some time ago... but perhaps not longer...

MatthewSandfordImprobable: Mmm, this still seem like it is an issue (we have to wait for worker flags to be set).

jeffxu-improbable: `CheckCmdLineParameters` only effects on the first time. 
Is there any chance when the first time it was called and the worker flags didn't came.
In this case, wait for seconds will not change anything. 

If we won't replace worker flags with command line arguments, perhaps we need find a way to make sure we wait worker flag values before we start to play.

MatthewSandfordImprobable: The issue this is fixing is that it is possible for CheckCmdLineParameters to never be called on a server (this only really happens in local testing). This can happen when FindPlayerStart is not called due to no attempt to spawn a player being made. 

It appears we always call FindPlayerStart after worker flags are set. This is really by chance and you are right that we should have a better way to deal with this. I think ideally we would cache player spawn requests and then action them once we receive the worker flags. Probably out of scope for this PR. I think I might add callback for when the worker values change though which would be neater.

MatthewSandfordImprobable: @jeffxu-improbable, I have updated the code to attempt to create player start actors only once updates from worker flags are received. Hopefully this is a little neater.

jeffxu-improbable: Hi @MatthewSandfordImprobable , sorry about reply late, too busy today. I will check it tomorrow. 
As we changed a lot of uptime game mode, it is better to have a uptime test scenario. I think `gdk-7duptime-zone2-1-manual.yaml` would be enough.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+	{
+		SpawnPointActor->Destroy();
+	}
+}
 
-	static ConstructorHelpers::FClassFinder<AActor> DropCubeClassFinder(TEXT("/Game/Benchmark/DropCube_BP"));
-	DropCubeClass = DropCubeClassFinder.Class;
+void USpawnCluster::GenerateSpawnPoints()
+{
+	const int32 DistBetweenSpawnPoints = 300;
```
simonsarginson, line 132: why 300?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
-			FVector PointA = PlayerStream.VRand()*RoamRadius;
-			FVector PointB = PlayerStream.VRand()*RoamRadius;
-			PointA.Z = PointB.Z = 0.0f;
-			PlayerRunPoints.Emplace(FBlackboardValues{ PointA, PointB });
-		}
+		const FVector& SpawnPoint = SpawnPoints[i];
+
+		FActorSpawnParameters SpawnInfo;
+		SpawnInfo.Owner = World->GetAuthGameMode();
+		SpawnInfo.Instigator = NULL;
```
simonsarginson, line 162: Nullptr?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+		SpawnInfo.bDeferConstruction = false;
+		SpawnInfo.SpawnCollisionHandlingOverride = ESpawnActorCollisionHandlingMethod::AlwaysSpawn;
+
+		// Swapping X and Y as GridBaseLBStrategy has the two reversed.
+		FVector ModifiedSpawnPoint = SpawnPoint;
+		ModifiedSpawnPoint.X = SpawnPoint.Y;
+		ModifiedSpawnPoint.Y = SpawnPoint.X;
+		ModifiedSpawnPoint.Z = 300;
```
simonsarginson, line 170: why 300?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
 	{
-		FRandomStream NPCStream;
-		NPCStream.Initialize(FCrc::MemCrc32(&TotalNPCs, sizeof(TotalNPCs)));
-		for (int i = 0; i < TotalNPCs; i++)
+		return;
+	}
+
+	// Sort the clusters by "distance from centre of area" ascending.
+	// This ensures we iterate from closest to furthest cluster.
+	ClusterPoints.Sort([this](const FVector& Point1, const FVector& Point2)
```
simonsarginson, line 192: worth breaking out into a free function?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+AActor* USpawnManager::GetSpawnPointActorByIndex(const int32 Index) const
 {
-	ClearExistingSpawnPoints();
-	GenerateSpawnPointClusters(NumPlayerClusters);
+	const int32 NumSpawnPoints = GetNumSpawnPoints();
+	if (NumSpawnPoints == 0)
+	{
+		return nullptr;
+	}
+	return SpawnPointActors[Index % NumSpawnPoints];
```
simonsarginson, line 251: Why do we want to wrap here?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+{
+	SpawnAreas.Empty();
+	SpawnPointActors.Empty();
+}
+
+void USpawnManager::GenerateSpawnAreas(const int32 ZoneRows, const int32 ZoneCols, const int32 ZoneWidth, const int32 ZoneHeight,
+	const int32 ZoneClusters, const int32 BoundaryClusters,
+	const int32 MaxSpawnPointsPerCluster, const int32 MinDistanceBetweenClusters)
+{
+	const float StartX = ZoneWidth / 2.0f - ZoneCols * ZoneWidth / 2.0f;
```
simonsarginson, line 269: I would suggest adding a comment here.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+				continue;
+			}
+
+			USpawnArea* NewSpawnArea = NewObject<USpawnArea>(this);
+			if (NewSpawnArea != nullptr)
+			{
+				NewSpawnArea->Type = bIsZone ? USpawnArea::Type::Zone : USpawnArea::Type::Boundary;
+
+				const float X = StartX + Col * ZoneWidth / 2;
+				const float Y = StartY + Row * ZoneHeight / 2;
```
simonsarginson, line 308:
```suggestion
				const float X = StartX + Col * ZoneWidth / 2.0f;
				const float Y = StartY + Row * ZoneHeight / 2.0f;
```
Minor

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+}
 
-	SpawnNPCs(TotalNPCs);
+// --- ABenchmarkGymGameMode ---
+
+ABenchmarkGymGameMode::ABenchmarkGymGameMode()
+	: DistBetweenClusterCenters(40000) // 400 meters, in Unreal units.
+	, PercentageSpawnPointsOnWorkerBoundaries(0.25f)
+	, bHasCreatedSpawnPoints(false)
+	, PlayerDensity(-1)
```
simonsarginson, line 347: What is PlayerDensity -1

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,366 @@
...
+	, PlayersSpawned(0)
+	, NPCSToSpawn(0)
+{
+	PrimaryActorTick.bCanEverTick = true;
+
+	static ConstructorHelpers::FClassFinder<AActor> DropCubeClassFinder(TEXT("/Game/Benchmark/DropCube_BP"));
```
simonsarginson, line 353: would error log here if we can't find the DropCube_BP

simonsarginson: You can use `DropCubeClassFinder.Succeeded()` for a conditional log

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -129,191 +399,115 @@ void ABenchmarkGymGameMode::Tick(float DeltaSeconds)
...
+	{
+		PlayerDensity = FCString::Atoi(*PlayerDensityString);
+	}
 }
 
-void ABenchmarkGymGameMode::ClearExistingSpawnPoints()
+void ABenchmarkGymGameMode::GenerateTestScenarioLocations()
 {
-	for (AActor* SpawnPoint : SpawnPoints)
+	constexpr float RoamRadius = 7500.0f; // Set to half the NetCullDistance
```
simonsarginson, line 431: What if someone tweak NetCullDistance in the editor, is that an issue?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: For some reason unreal natively initializes it as NULL. Not sure why. I have remove the setting of SpawnInfo.Instigator as it is redundant.

simonsarginson: Probably legacy engine code from unreal 3 days or even earlier.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: There are only so many spawn point actors (NumSpawnPoints). We always want to return a valid spawn point actor so we looop.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: Added a comment but essentially it is invalid until it is set.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: Breaking GenerateSpawnClusters into a free function?

simonsarginson: The lambda, you use it another function too.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
simonsarginson: So what is the point on the index, why not just give out the next spawnpoint?

m-samiec:
Could you clarify any changes made to blueprint assets? 🙏

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
```diff
@@ -2,433 +2,104 @@
...
-	, TestDataFrequency(0)
+	: EgressSize(0)
+	, EgressFrequency(0)
 	, CrossServerSize(0)
 	, CrossServerFrequency(0)
-	, NumPlayerClusters(1)
-	, PlayersSpawned(0)
-	, NPCSToSpawn(0)
 {
+	PercentageSpawnPointsOnWorkerBoundaries = 0.0f;
```
m-samiec, line 31: Initialise this in the class it's declared in, not here

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
```diff
@@ -447,37 +118,60 @@ void AUptimeGameMode::SpawnCrossServerActors(int32 CrossServerPointNum)
...
-	const float DistBetweenRows = ZoneHeight / SpawnRows;
-	const float DistBetweenCols = ZoneWidth / SpawnCols;
-	float StartingX = -(SpawnCols - 1) * ZoneWidth / 2 / SpawnCols;
-	float StartingY = -(SpawnRows - 1) * ZoneHeight / 2 / SpawnRows;
+	const float Height = GetZoneHeight();
+	const float Width = GetZoneWidth();
+	const float Rows = GetZoningRows();
+	const float Cols = GetZoningCols();
+
+	const float DistBetweenRows = Height / Rows;
```
m-samiec, line 134: nit: semantics, but if we're dividing than that would suggest this is the WorldHeight, not the ZoneHeight

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -174,6 +175,39 @@ void ABenchmarkGymGameModeBase::InitialiseActorCountCheckTimer()
 	}
 }
 
+void ABenchmarkGymGameModeBase::GatherWorkerConfiguration()
+{
+	// No need to fiddle with configuration as the defaults should reflect the native scenario.
```
m-samiec, line 180: default should reflect the "single server scenario which is all that's required in native" ?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -174,6 +175,39 @@ void ABenchmarkGymGameModeBase::InitialiseActorCountCheckTimer()
...
+		return;
+	}
+
+	const UWorld* World = GetWorld();
+	const UAbstractSpatialMultiWorkerSettings* MultiWorkerSettings =
+		USpatialStatics::GetSpatialMultiWorkerClass(World)->GetDefaultObject<UAbstractSpatialMultiWorkerSettings>();
+
+	if (MultiWorkerSettings != nullptr && MultiWorkerSettings->WorkerLayers.Num() > 0)
+	{
+		NumWorkers = 0;
```
m-samiec, line 192: Can just call MultiWorkerSettings->GetMinimumRequiredWorkerCount() here, it'll do the iteration internally

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
m-samiec: but I see that ZoneHeight is actually the zone height - hmm
so we take the world height - divide by rows to get zone height, then divide by rows again to get DistBetweenRows?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,373 @@
...
 
 DEFINE_LOG_CATEGORY(LogBenchmarkGymGameMode);
 
 namespace
 {
 	const FString PlayerDensityWorkerFlag = TEXT("player_density");
 	const FString BenchmarkPlayerDensityCommandLineKey = TEXT("PlayerDensity=");
-	const float PercentageSpawnpointsOnWorkerBoundary = 0.25f;
+
+	// Returns an array of co-ordinates. Each co-ordinate represents the center of a cell in a generated grid.
```
m-samiec, line 31:
```suggestion
	// Returns an array of coordinates. Each coordinate represents the center of a cell in a generated grid.
```

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -19,96 +19,373 @@
...
-		NPCStream.Initialize(FCrc::MemCrc32(&TotalNPCs, sizeof(TotalNPCs)));
-		for (int i = 0; i < TotalNPCs; i++)
+		return;
+	}
+
+	// Sort the clusters by "distance from centre of area" ascending.
+	// This ensures we iterate from closest to furthest cluster.
+	ClusterPoints.Sort([this](const FVector& Point1, const FVector& Point2)
+	{
+		const float Distance1 = FVector::Distance(Point1, WorldPosition);
```
m-samiec, line 195: uber unnecessary performance nit here - but you can use DistSquared to prevent the sqrt cost.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: POOOOG!

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
MatthewSandfordImprobable: Actually this might be mistake. Previously we divided by SpawnRows/SpawnCols. These were set by worker flags. It may be the case these simply lined up with the number of ZoneRows / ZoneCols but I can't remember so will have a look.

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
MatthewSandfordImprobable: So this is the derived class (AUptimeGameMode) overriding the value. Is there a nice way to do this?

MatthewSandfordImprobable:
> Could you clarify any changes made to blueprint assets? 🙏

No "changes" but some additional assets that help configure a 2x2 scenario.

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
MatthewSandfordImprobable: Yeah taking a look at what we have in TestGymBuildkite, SpawnCols and SpawnRows just lined up with how how many zone rows and zone columns there were in the test. This logic seems a little funky... I was just trying to reproduce what was there but... I don't think it makes sense for us to divide by the SpawnCols/SpawnRows (ZoningRows/ZoningCols) twice as you pointed out. I'll take a look at how to do this in a nicer way.

MatthewSandfordImprobable: Think I am overcomplicating thing... think we simply want one beacon spawned in the middle of each zone...

MatthewSandfordImprobable:
> Could you clarify any changes made to blueprint assets? 🙏

There are no "changes" only additions. The two assets added are to configure a 2x2 zoning strategy.

MatthewSandfordImprobable:
> > Could you clarify any changes made to blueprint assets? 🙏
> 
> No "changes" but some additional assets that help configure a 2x2 scenario.

Ok somewhat of a lie now. I added a driveby fix for uptime player character. They were not spawning drop cubes in the uptime scenario,

Game/Source/GDKTestGyms/AUptimeGameMode.cpp
```diff
@@ -2,482 +2,159 @@
...
-		UE_LOG(LogUptimeGymGameMode, Error, TEXT("Error spawning NPC, World is null"));
+		UE_LOG(LogUptimeGymGameMode, Error, TEXT("Error spawning, SpawnManager is null"));
 		return;
 	}
 
-	if (NPCSpawner == nullptr)
-	{
-		ABenchmarkGymNPCSpawner* Spawner = nullptr;
-		for (TActorIterator<ABenchmarkGymNPCSpawner> It(GetWorld()); It; ++It)
+	SpawnManager->ForEachZoneArea([this, World](USpawnArea& ZoneArea) {
```
m-samiec, line 120: nit - could make this `ForEachArea` and have the user pass in an enum to match the type of area they want to execute on

Game/Source/GDKTestGyms/AUptimeGameMode.h
```diff
@@ -2,91 +2,54 @@
...
-	TArray<ControllerIntegerPair> AIControlledPlayers;
+	virtual void ReadCommandLineArgs(const FString& CommandLine) override;
+	virtual void ReadWorkerFlagValues(USpatialWorkerFlags* SpatialWorkerFlags) override;
+	virtual void BindWorkerFlagDelegates(USpatialWorkerFlags* SpatialWorkerFlags) override;
+
+private:
 
-	virtual void Tick(float DeltaSeconds) override;
-	virtual void ParsePassedValues() override;
+	int32 EgressSize;
```
m-samiec, line 35: EgressSize and EgressFrequency don't actually appear to be used atm - are these available for some WIP functionality?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.h
```diff
@@ -10,11 +10,92 @@
...
+	float Height;
+	int32 MaxSpawnPoints;
+	float MinDistanceBetweenSpawnPoints;
+
+	bool GenerateSpawnPoints(const TArray<AActor*>** OutSpawnPointActors);
+
+private:
+
+	UPROPERTY()
+	TArray<AActor*> SpawnPointActors;
```
m-samiec, line 34: Should this be <APlayerStart*> ?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.h
```diff
@@ -10,11 +10,92 @@
...
+	bool GenerateSpawnClusters(const TArray<AActor*>** OutSpawnPointActors);
+	AActor* GetSpawnPointActorByIndex(const int32 Index) const;
+
+private:
+
+	UPROPERTY()
+	TArray<USpawnCluster*> SpawnClusters;
+
+	UPROPERTY()
+	TArray<AActor*> SpawnPointActors;
```
m-samiec, line 67: also here

simonsarginson (approved):

Game/Source/GDKTestGyms/AUptimeGameMode.h
MatthewSandfordImprobable: AUptimeGymsCharacter pulls the values using getters in the header.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
MatthewSandfordImprobable: You mean have an internal counter that tracks what the next spawn point is? If so this reduces the utility of the function, utility that we actually do use. One point in the code specifically asks each spawn area for a spawn actor at index 0 (the center).

m-samiec (approved):

PR #427, opened 2021-06-23
Change game mode on WorldCompositionGym map to default

MatthewSandfordImprobable:
WorldCompositionGym map was using a game mode that it did not need to so I have changed it to use the default game mode.

The exception was caused by an actor class not being defined on the game mode. I will put a check for a valid class in a PR that I have open at the moment.

image


joshuahuburn (approved):

m-samiec:
Doesn't this need to go into the rc?

MatthewSandfordImprobable:
> Doesn't this need to go into the rc?

Sweet will retarget.

PR #433, opened 2021-06-30
Update Gameplay Debugger gym steps

oblm:
I'm testing the gameplay debugger. I'm clarifying the testing steps as I go.


oblm:
Abandoned this PR as, while remapping works for `num4` in this context, I cannot remap `num*`, which makes it an incomplete and inadequate solution.

If `num*` is used by Epic's native debugger I will open a feature request for them to add a remap option for that to this window: 
image

If not, then we may want to reconsider remapping our use of  `num*` to a remappable key like `num9`, which appears unused.

PR #437, opened 2021-07-05
[UNR-5791] Gameplay Debugger tracks focused actor across interest boundaries

martin-improbable:
Test gym changed to "no-interest" (from "full-interest"), and the test plan verifies that "no-interest" tracking works as expected.


USER_MANUAL.md
```diff
@@ -570,6 +571,10 @@ The late connecing client has validated the local state before sending the "Pass
   * Press `ctrl+numpad+` or `ctrl+numpad-` to cycle to the next server.
   * Observe that the "eqs debug spheres" are drawn and updated only within the newly selected server.
   * Click the `Stop` button in the Unreal toolbar to stop the PIE session
+  * Click the `Play` button in the Unreal toolbar to start PIE session, check the option for `SpatialOS networking`, and set `Number of players` to 1.
+  * Keeping the player within the server in which it started, before the `GameplayDebuggerGymCube` reaches the server boundary for the first time press `'`(Apostrophe) to enable the gameplay debugger whilst looking directly at the cube.
+  * Still keeping the player within the server in which it started, observe there are "EQS debug spheres" drawn at regular intervals around the cube as it moves along its path, within the server boundary in which the player is positioned.
```
improbable-fabian, line 576: I think one failure case to watch out for is that the cube moves to the other zone (and our debug spheres remain in place because our manually selected debug server is no longer authoritative over it) but when it moves back to our zone they aren't updated anymore because the server "lost track of it". This should now work properly even in a no interest setup, where as long as you have the debug actor selected you'll never lose it. Should we capture this here?

improbable-fabian (approved):
LGTM % potentially one small addition

PR #449, opened 2021-07-13
Fix checks for multiple actor counts

m-samiec:
Previously we were allowing failures to go unreported if they weren't the last actor count checked in Spatial runs.

### DIFF
--- a/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
+++ b/Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
@@ -959,6 +959,7 @@ void ABenchmarkGymGameModeBase::UpdateAndCheckTotalActorCounts()
 
 	UE_LOG(LogBenchmarkGymGameModeBase, Log, TEXT("--- Actor Count Totals ---"));
 
+	bActorCountFailureState = false; // Reset check.
 	for (const auto& ActorCountPair : TempTotalActorCounts)
 	{
 		const TSubclassOf<AActor>& ActorClass = ActorCountPair.Key;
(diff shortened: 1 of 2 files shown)


Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -972,16 +973,20 @@ void ABenchmarkGymGameModeBase::UpdateAndCheckTotalActorCounts()
...
 			{
-				bHasActorCountFailed = true;
-				NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Unreal actor count check. ObjectClass %s, MinCount %d, MaxCount %d, ActualCount %d"),
-					*NFRFailureString,
-					*ActorClass->GetName(),
-					ExpectedActorCount.MinCount,
-					ExpectedActorCount.MaxCount,
-					TotalActorCount);
+				bActorCountFailureState = true;
+				if (!bHasActorCountFailed)
```
simonsarginson, line 980: Minor: `bLoggedActorCountFailed`
Quite difficult to read with so many similar names.
Alternatively could we just return after logging instead?

m-samiec: Will rework to remove `bActorCountFailed`
Not sure what returning early would help with though

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -972,16 +973,20 @@ void ABenchmarkGymGameModeBase::UpdateAndCheckTotalActorCounts()
 		{
 			// Check for test failure
 			const FExpectedActorCountConfig& ExpectedActorCount = ExpectedActorCounts[ActorClass];
-			bActorCountFailureState = TotalActorCount < ExpectedActorCount.MinCount || TotalActorCount > ExpectedActorCount.MaxCount;
-			if (bActorCountFailureState && !bHasActorCountFailed)
+			bool bActorCountFailed = TotalActorCount < ExpectedActorCount.MinCount || TotalActorCount > ExpectedActorCount.MaxCount;
```
MatthewSandfordImprobable, line 976: As an alternative to reduce the number of similarly named variabled... we might be able to reset bActorCountFailureState to false as you do but do:

bActorCountFailureState |=TotalActorCount < ExpectedActorCount.MinCount || TotalActorCount > ExpectedActorCount.MaxCount;

Think this should work and keep the rest of the code the same?

m-samiec: Reworked from @simonsarginson comments already - pt another look

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
simonsarginson: Maybe I am reading it wrong but it seems we can only log "Unreal actor count check" once. So rather than using a bool to gate that this logs only once, just return instead?
```
...
	for (const auto& ActorCountPair : TempTotalActorCounts)
	{
		...
		if (bIsReadyToConsiderActorCount)
		{
			// Check for test failure
			const FExpectedActorCountConfig& ExpectedActorCount = ExpectedActorCounts[ActorClass];
			if (TotalActorCount < ExpectedActorCount.MinCount || TotalActorCount > ExpectedActorCount.MaxCount)
			{
				bActorCountFailureState = true;
				bHasActorCountFailed = true;
				NFR_LOG(LogBenchmarkGymGameModeBase, Error, TEXT("%s: Unreal actor count check. ObjectClass %s, MinCount %d, MaxCount %d, ActualCount %d"),
						*NFRFailureString,
						*ActorClass->GetName(),
						ExpectedActorCount.MinCount,
						ExpectedActorCount.MaxCount,
						TotalActorCount);
				return;
			}
		}

m-samiec: We want `bActorCountFailureState` to be the current state of failre, and the `bHasActorCountFailed` to store if the test has ever failed. So we need to continue to do the check regardless of the current state of `bHasActorCountFailed`

m-samiec: And yes - to prevent log spam - we only output the failure message once.

MatthewSandfordImprobable (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
simonsarginson: Beyond the scope of this ticket but I'd love to see this broken up a bit in a few different functions, its hard work reading this, but maybe its just me.

simonsarginson (approved):

PR #450, opened 2021-07-14
[UNR-5775] Fix actor count metric being incorrectly reported

MatthewSandfordImprobable:

PR #458, opened 2021-07-30
Re-add actor count and revert recent AlwaysReplicate change

m-samiec:
Reverting the rep graph changes introduced here (https://github.com/spatialos/UnrealGDKTestGyms/pull/450) and added actor count logging.


mironec (approved):
Probably good, but not sure what the uasset changes are doing

Game/Source/GDKTestGyms/Private/CounterComponent.cpp
```diff
@@ -0,0 +1,52 @@
...
+		UpdateCachedActorCounts();
+	}
+}
+
+int32 UCounterComponent::GetActorClassCount(TSubclassOf<AActor> ActorClass) const
+{
+	const int32* CachedCount = CachedClassCounts.Find(ActorClass);
+	if (CachedCount == nullptr)
+	{
+		UE_LOG(LogTemp, Warning, TEXT("Could not get actor class count for class %s"), *ActorClass->GetName());
```
mironec, line 33: LogCounterComponent?

Game/Source/GDKTestGyms/Private/CounterComponent.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+	Timer -= DeltaTime;
+	if (Timer <= 0.0f)
+	{
+		Timer = ReportFrequency;
+		UpdateCachedActorCounts();
+	}
+}
+
+int32 UCounterComponent::GetActorClassCount(TSubclassOf<AActor> ActorClass) const
```
mironec, line 28: is this function even used? wouldn't we just report the current situation in update? as in - do we need the caching?

m-samiec: You're right this isn't currently used, but this is just reverting files that already existed, so will leave for the time being.

m-samiec:
> Probably good, but not sure what the uasset changes are doing

Sorry should have specified - adding the Count Component back to the Game States, and outputting the counts of the actors we care about.

PR #459, opened 2021-08-09
Fix for nonspatialized classes in Rep Graph

m-samiec:
Required PR for https://github.com/spatialos/UnrealGDK/pull/3285

### DIFF
--- a/Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
+++ b/Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
@@ -193,6 +193,10 @@ void UTestGymsReplicationGraph::InitGlobalActorClassSettings()
 		{
 			AddInfo(Class, EClassRepNodeMapping::RelevantAllConnections);
 		}
+		else if (bUsingSpatial && ActorCDO->GetIsReplicated())
+		{
+			AddInfo(Class, EClassRepNodeMapping::AlwaysReplicate);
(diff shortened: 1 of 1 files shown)


Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -193,6 +193,10 @@ void UTestGymsReplicationGraph::InitGlobalActorClassSettings()
 		{
 			AddInfo(Class, EClassRepNodeMapping::RelevantAllConnections);
 		}
+		else if (bUsingSpatial && ActorCDO->GetIsReplicated())
+		{
+			AddInfo(Class, EClassRepNodeMapping::AlwaysReplicate);
+		}
```
mironec, line 199: Hm, sorry, I don't understand - do we fall through to this case for the actors from the other PR?
Does this also trigger for any other classes?

m-samiec: Yep, previously classes here would have fallen into the `Not adding info for class` section. This would mean that actor's of that type would default to any parent's class routing, which is not what we want. So `OnlyRelevantToOwner` etc. actors were being treated as Dynamic actors. This now just ensures from the server's side that we just replicate the actor, and we let QBI deal with the interest side.

mironec (approved):

PR #462, opened 2021-08-31
Fix player spawning clustering

m-samiec:
Whilst investigating https://github.com/spatialos/UnrealGDK/pull/3301 found that using `GetUniqueID` wasn't valid in a multi-server context where Player Actors were being destroyed and reused. This meant that spawning wasn't equal, and some cluster were getting >8 players.

Also includes a driveby for the GameState to also count the numbers of PlayerStates, which *should* be equal to number of players.


Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -613,18 +613,9 @@ AActor* ABenchmarkGymGameMode::FindPlayerStart_Implementation(AController* Playe
 		return SpawnManager->GetSpawnPointActorByIndex(PlayersSpawned);
 	}
 
-	// Use custom spawning with density controls
-	const int32 PlayerUniqueID = Player->GetUniqueID();
-	AActor** SpawnPoint = PlayerIdToSpawnPointMap.Find(PlayerUniqueID);
```
improbable-danny, line 618: Can `PlayerIdToSpawnPointMap` be removed completely?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -613,18 +613,9 @@ AActor* ABenchmarkGymGameMode::FindPlayerStart_Implementation(AController* Playe
...
-	if (SpawnPoint != nullptr)
-	{
-		return *SpawnPoint;
```
improbable-danny, line 621: iirc FindPlayerStart_Implementation is called by SpatialSpawner to determine worker zone and then again during actual spawning logic, is this still a problem?

improbable-danny (approved):

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
m-samiec: Yeah it is, but sets Player to nullptr to indicate it isn't a "real" player spawning instance. And we special case for that just above this code here.

improbable-danny (approved):

PR #469, opened 2021-09-30
[UNR-5578] multiworker gym cube jitter

iwanmehanda:
Increased the minimum net update frequency  on NoNetLoadGymCube to make sure it is replicated at every server tick. This doesn't remove the jittering effect, but can help minimize it:

image

Also updated the User Manual of this Test Gym to better reflect the findings.

**Explanation**:
The jittering effect reported seems to be a visual artefact and not a functional one, caused by a framerate mismatch. The Gyms have a max server tick rate of 30 FPS while clients play on 60 FPS by default. The cubes will stop jittering when the client FPS is capped by running the console command `t.MaxFPS 30`. We can verify the cube is not functionally jittering because a character standing on it would not see the rest of the map jitter.

After some tinkering with the `TestGymReplicationGraph` class and enabling/disabling it in the `DefaultEngine.ini`, it does not have any obvious effect on the severity of the jittering, so the issue is likely unrelated.


martin-improbable (approved):

simonsarginson (approved):

simonsarginson:
LGTM!

PR #475, opened 2021-11-17
Only create latency tracer when needed

MatthewSandfordImprobable:
The latency tracer on the game instance will now only be created and initialized when explicitly asked for. We were creating tracers in all GDKTest gyms maps. This also lead to empty etlog files being generated. This change also means we don't have an unused tracer sitting around in many of our test scenarios.


improbable-danny (approved):

simonsarginson (approved):
LGTM, nice work!

m-samiec (approved):

PR #476, opened 2021-11-19
[UNR-5654] Collision handover gym fix

biancasapu:
The gym had an issue where the cubes would jankily pass through the floor, even if collision was disabled through the blueprint. The PR fixes this issue, and the cubes cleanly fall through the floor, including over handover.

The original blueprint for the CollisionCube actor:
collisioncube-before

For 1 connected client, `SetActorEnableCollision`  was called as such:
- WorkerRegion 0 disables collision (ROLE_Authority)
- WorkerRegion 1 disables collision  (ROLE_Authority)
- Cube 0 disables collision (ROLE_Authority)
- Cube 1 disables collision  (ROLE_Authority)
- Cube 0 disables collision (ROLE_SimulatedProxy)
- Cube 1 disables collision  (ROLE_SimulatedProxy)

Removing the branch and AlreadyBegunPlay bool:
collisioncube-after

After the change, 
- WorkerRegion 0 disables collision (ROLE_Authority)
- WorkerRegion 1 disables collision  (ROLE_Authority)
- Cube 0 disables collision (ROLE_Authority)
- Cube 1 disables collision  (ROLE_Authority)
- Cube 0 disables collision (ROLE_SimulatedProxy)
- Cube 1 disables collision  (ROLE_SimulatedProxy)
- Cube 0 disables collision (ROLE_SimulatedProxy) (again)
- Cube 1 disables collision (ROLE_SimulatedProxy) (again)

In order to check that handover actually works, we won't disable collisions on the non-auth server. We need to have it disabled on the non-auth clients though, since the client has no way of knowing collision was turned off by the server. Also, if the client tries to disable collisions while the server still has them enabled, collision will stay enabled.

The final blueprint is:
collisioncube-final

And now we can watch the cubes fall smoothly 🎉


mironec:
Can't remember exactly how the gym is set up, but maybe we want the second server to have collision enabled at first? So that we can verify the collision has been handed over between servers?
Otherwise, seems good to me.
(Also reading this reminded me why we did it in the first place, and I am pretty sure I just want to remove the handover of collision)

joshuahuburn (approved):

mironec (approved):

PR #477, opened 2021-11-22
Feature/enable entity id interest

m-samiec:
REVIEW IN COMMIT ORDER

Full integration of USI, as well as a range of small functionality improvements that were made along the way.

Blueprint changes are all minimal config changes.

GDK - https://github.com/spatialos/UnrealGDK/pull/3339


Game/Source/GDKTestGyms/Private/UserExperienceReporter.cpp
```diff
@@ -8,7 +8,9 @@
 #include "Net/UnrealNetwork.h"
 #include "EngineClasses/SpatialNetDriver.h"
 #include "UserExperienceComponent.h"
+#include "Interop/Connection/SpatialWorkerConnection.h"
 #include "Utils/SpatialMetrics.h"
+#include "Utils/SpatialStatics.h"
```
MatthewSandfordImprobable, line 13: What is SpatialStatics.h used for?

Game/Source/GDKTestGyms/Private/UserExperienceComponent.cpp
```diff
@@ -32,10 +32,11 @@ void UUserExperienceComponent::StartRoundtrip()
 {
 	if (OpenRPCs.Contains(RequestKey))
 	{
+		UE_LOG(LogTemp, Warning, TEXT("UserExperience failure: Previous round trip didn't complete in time (%d)."), RequestKey);
```
MatthewSandfordImprobable, line 35: If this is a "failure", will we fail NFRs based on the presence of this log?

Game/Source/GDKTestGyms/Private/UserExperienceComponent.cpp
```diff
@@ -96,6 +100,16 @@ float UUserExperienceComponent::CalculateAverageUpdateTimeDelta() const
 	return Avg;
 }
 
+void UUserExperienceComponent::CheckPosition()
+{
+	FVector CurrPos = GetOwner()->GetActorLocation();
+	if (PreviousPos == CurrPos)
+	{
+		UE_LOG(LogTemp, Warning, TEXT("UserExperience failure: Client position not updating (%s). Vel %s"), *PreviousPos.ToString(), *GetOwner()->GetVelocity().ToString());
```
MatthewSandfordImprobable, line 108: Similarly, will we fail NFRs based on the presence of this log?

Game/Source/GDKTestGyms/Private/LatencyTracer.cpp
```diff
@@ -3,7 +3,13 @@
 #include "LatencyTracer.h"
 
 #include "Engine/World.h"
+
+#include "EngineClasses/SpatialNetDriver.h"
```
MatthewSandfordImprobable, line 7: Thank you!

Game/Source/GDKTestGyms/Public/TestGymsCharacterMovementComp.h
```diff
@@ -23,7 +23,7 @@ class GDKTESTGYMS_API UTestGymsCharacterMovementComp : public UCharacterMovement
 	void ClientAuthServerMove_Implementation(const FVector_NetQuantize100& ClientLocation, const FVector_NetQuantize10& ClientVelocity, const uint32 PackedPitchYaw);
 	
 	bool bClientAuthMovement = false;
-	int32 ClientMoveFrequency = 10; // Send client position updates to server at this frequency
+	int32 ClientMoveFrequency = 30; // Send client position updates to server at this frequency
```
MatthewSandfordImprobable, line 26: Guess this affects all many of our tests. What was the purpose here?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -899,6 +890,40 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayerMovement_Implementation
...
+
+	if (!WorkerID.IsEmpty())
+	{
+		LatestClientUXMap.Emplace(WorkerID, UX{RTTime, UpdateTime});
+	}
+
+	AveragedClientRTTMS = 0.f;
+	AveragedClientUpdateTimeDeltaMS = 0.f;
+
+	for (auto& Entry : LatestClientUXMap)
```
MatthewSandfordImprobable, line 914: const auto& Entry

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -899,6 +890,40 @@ void ABenchmarkGymGameModeBase::ReportAuthoritativePlayerMovement_Implementation
...
+	{
+		AveragedClientRTTMS = FMath::Max(AveragedClientRTTMS, Entry.Value.RTT);
```
MatthewSandfordImprobable, line 916: I am probably missing something here but... AveragedClientRTTMS seems to be set to the maximum of all workers ClientRTTMS instead of the average?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -147,6 +147,15 @@ void ABenchmarkGymGameModeBase::BeginPlay()
 	{
 		OutputPlayerDensity();
 	}
+
+	if (bLongFormScenario)
+	{
+		// Extend timers to handle longer expected deployment lifetime (required for current long form disco performance test)
+		RequiredPlayerCheckTimer.SetTimer(17 * 60);
+		DeploymentValidTimer.SetTimer(38 * 60);
+		UNFRConstants* NFRConstants = const_cast<UNFRConstants*>(UNFRConstants::Get(GetWorld()));
```
MatthewSandfordImprobable, line 156: Feel as if there could be room for a mini rework of how UNFRConstants work but... maybe in future.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -488,6 +488,7 @@ void ABenchmarkGymGameMode::GenerateTestScenarioLocations()
 {
 	const APawn* Pawn = GetDefault<APawn>(SimulatedPawnClass);
 	const float RoamRadius = FMath::Sqrt(Pawn->NetCullDistanceSquared) / 2.0f;
+	const float MinDistanceSq = 500.f * 500.f;	// Move To threshold used in behaviour tree
```
MatthewSandfordImprobable, line 491: Nice catch!

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -496,6 +497,10 @@ void ABenchmarkGymGameMode::GenerateTestScenarioLocations()
 			FVector PointA = PlayerStream.VRand()*RoamRadius;
 			FVector PointB = PlayerStream.VRand()*RoamRadius;
 			PointA.Z = PointB.Z = 0.0f;
+			if (FVector::DistSquared2D(PointA, PointB) < MinDistanceSq)
+			{
+				PointA *= -1.f;
```
MatthewSandfordImprobable, line 502: Does this have the possibility of causing characters to move out of the "density regions"?
Perhaps we just generate random points till we find a valid pair?

MatthewSandfordImprobable (commented):
Would you be able to give a quick sum up of what happened in the blueprints? Looks like the component was broken before these changes? 

image

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -749,6 +880,35 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::GatherActorListsForConnecti
 	Params.OutGatheredReplicationLists.AddReplicationActorList(ReplicationActorList);
 }
 
+void UTestGymsReplicationGraphNode_GlobalViewTarget::GatherClientInterestedActors(const FConnectionGatherActorListParameters& Params)
+{
+	ReplicationActorList.Reset();
+
+	for (const FNetViewer& CurViewer : Params.Viewers)
+	{
+		if (CurViewer.InViewer)
```
MatthewSandfordImprobable, line 889: Some explicit checks here

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -757,6 +917,87 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::LogNode(FReplicationGraphDe
 	DebugInfo.PopIndent();
 }
 
+
+void UTestGymsReplicationGraphNode_NearestActors::NotifyAddNetworkActor(const FNewReplicatedActorInfo& ActorInfo)
+{
+	ReplicationActorList.Add(ActorInfo.Actor);
+}
+
+bool UTestGymsReplicationGraphNode_NearestActors::NotifyRemoveNetworkActor(const FNewReplicatedActorInfo& ActorInfo, bool bWarnIfNotFound /*= true*/)
```
MatthewSandfordImprobable, line 926: Maybe we should warn if not found actor to remove?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -808,3 +1049,22 @@ FAutoConsoleCommandWithWorldAndArgs ChangeFrequencyBucketsCmd(TEXT("TestGymsRepG
 		Node->SetNonStreamingCollectionSize(Buckets);
 	}
 }));
+
+FAutoConsoleCommandWithWorldAndArgs ChangeDensityCmd(TEXT("TestGymsRepGraph.AlterNearestN"), TEXT("Alters nearest actor density"),
```
MatthewSandfordImprobable, line 1053: Coool!

Game/Source/GDKTestGyms/Private/UserExperienceReporter.cpp
m-samiec: Removed

Game/Source/GDKTestGyms/Private/UserExperienceComponent.cpp
m-samiec: No it won't, as the client logs aren't parsed. But this log would most likely indicate that the UX metric has failed, and having "UserExperience failure" is useful for bulk searching.

Game/Source/GDKTestGyms/Private/UserExperienceComponent.cpp
m-samiec: Same as above

Game/Source/GDKTestGyms/Public/TestGymsCharacterMovementComp.h
m-samiec: This is only used for client-auth movement rpcs, which are only used in the new disco scenario.

m-samiec: And purpose was to make the scenario more "representative"

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
```diff
@@ -614,7 +614,7 @@ AActor* ABenchmarkGymGameMode::FindPlayerStart_Implementation(AController* Playe
 {
 	if (SpawnManager->GetNumSpawnPoints() == 0)
 	{
-		return Super::FindPlayerStart_Implementation(Player, IncomingName);
+		return nullptr;
```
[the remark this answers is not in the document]
m-samiec: Yes, as the SpawnManager is only initialised on the auth game mode, if the spatial spawner attempts to access a local non-auth game mode, it should attempt to pass the message to the auth game mode. Previously when running in editor this was calling the super function, which was returning the WorldSettings obj which was used as a starting location, and the Spawn Point logic wasn't respected.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: Yeah I'll agree it's confusing. So each server worker averages it's client worker ux metrics, and sends that as a single value to the auth game mode. The auth game mode then looks for the worst aggregated average, from any one workers perspective, and if it's worse than the threshold, fails the test. I'll update the variable naming to try to make that clearer.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: Definitely agree

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
m-samiec: PointA and B are relative vectors of the same length (RoamRadius) that are added onto an origin position to generate the two waypoint coordinates. If we generated infinite points and plotted them, we'd get a perfect circle. This is just checking that if the two points that we've generated onto that circle, are too close to one another, flip the position of the first point (so that it's on the other side of the circle).

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
m-samiec: I don't think that's necessary, we return if the remove was successful or not, so the user of the node could warn if necessary.

m-samiec:
> Would you be able to give a quick sum up of what happened in the blueprints? Looks like the ReplicatedComponent was broken before these changes?
> 
> Also what was the counter object added to the game state?
> 
> image

Mentioned in the PR description that all blueprint changes are just config changes.

ReplicatedComponent_BP was just an option to replicate less data (this isn't enabled by default, but was useful in the stress testing experimentation). As all the characters+npcs use this component, it required re-saving them.

The DiscoGameState involved just added a CounterComponent similar to what's already setup on the BenchmarkGameState_BP class. This is just for outputting the at some frequency the count of specific actor counts into each worker's logs.

Game/Source/GDKTestGyms/Private/UserExperienceComponent.cpp
```diff
@@ -76,10 +77,13 @@ void UUserExperienceComponent::OnRep_ClientTimeTicks(int64 OldTicks)
...
 	Super::OnClientOwnershipGained();
+
 	FTimerHandle Timer;
-	FTimerDelegate Delegate;
-	Delegate.BindUObject(this, &UUserExperienceComponent::StartRoundtrip);
-	GetWorld()->GetTimerManager().SetTimer(Timer, Delegate, 1.0f, true);
+	GetWorld()->GetTimerManager().SetTimer(Timer, this, &UUserExperienceComponent::StartRoundtrip, 1.0f, true);
+
+	FTimerHandle PositionCheckTimer;
+	GetWorld()->GetTimerManager().SetTimer(PositionCheckTimer, this, &UUserExperienceComponent::CheckPosition, 10.0f, true, 10.0f);
```
simonsarginson, line 85: Is this safe, will this blow up if `UUserExperienceComponent` get gced from under it?

Game/Source/GDKTestGyms/Private/UserExperienceReporter.cpp
```diff
@@ -96,6 +98,10 @@ void UUserExperienceReporter::ReportMetrics()
 			if (Constants->ClientFPSMetricDelay.HasTimerGoneOff() && GameInstance->GetAveragedFPS() < Constants->GetMinClientFPS())
 			{
 				bFrameRateValid = false;
+				if (const USpatialNetDriver* NetDriver = Cast<USpatialNetDriver>(GameInstance->GetWorld()->GetNetDriver()))
+				{
+					UE_LOG(LogTemp, Error, TEXT("Client %s failed with fps %f below min threshold %f"), *NetDriver->Connection->GetWorkerId(), GameInstance->GetAveragedFPS(), Constants->GetMinClientFPS());
```
simonsarginson: `LogTemp` ?

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
simonsarginson: Worth a comment?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -730,6 +751,16 @@ void ABenchmarkGymGameModeBase::ReadWorkerFlagValues(USpatialWorkerFlags* Spatia
 		SetLifetime(FCString::Atoi(*LifetimeString));
 	}
 
+	if (SpatialWorkerFlags->GetWorkerFlag(CubeRespawnBaseTimeWorkerFlag, CubeRespawnBaseTimeString))
+	{
+		CubeRespawnBaseTime = FCString::Atof(*CubeRespawnBaseTimeString);
```
simonsarginson, line 756: This returns 0.0 on failure to make any sense of the string, is that acceptable?
(It calls `_tstof`  -> `_wtof` under the hood https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/atof-atof-l-wtof-wtof-l?view=msvc-170).

simonsarginson: Would actually suggest wrapping this in 
```
if (CubeRespawnBaseTimeString.IsNumeric())
{
}
// or
checkf(CubeRespawnBaseTimeString.IsNumeric(), TEXT("Not a valid number"));
```

simonsarginson: Might even be worth wrapping into a free function since there is a lot of string to float in this commit.

Game/Source/GDKTestGyms/BenchmarkGymGameMode.cpp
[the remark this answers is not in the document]
simonsarginson: worth a comment?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -149,6 +149,14 @@ void ABenchmarkGymGameModeBase::BeginPlay()
 	}
 }
 
+void ABenchmarkGymGameModeBase::OnAuthorityLost()
+{
+	// ABenchmarkGymGameModeBase doesn't support authority transfer
```
simonsarginson, line 163: Comment can be wrapped into a `ensuref` ?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -419,8 +427,7 @@ void ABenchmarkGymGameModeBase::Tick(float DeltaSeconds)
 	
 	// PrintMetricsTimer needs to be reset at the the end of ABenchmarkGymGameModeBase::Tick.
 	// This is so that the above function have a chance to run logic dependant on PrintMetricsTimer.HasTimerGoneOff().
-	if (HasAuthority() &&
-		PrintMetricsTimer.HasTimerGoneOff())
+	if (PrintMetricsTimer.HasTimerGoneOff())
```
simonsarginson, line 439: Why this change?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
```diff
@@ -99,11 +100,10 @@ ABenchmarkGymGameModeBase::ABenchmarkGymGameModeBase()
 	, UXAuthActorCount(0)
 	, PrintMetricsTimer(10)
 	, TestLifetimeTimer(0)
-	, TickActorCountTimer(60) // 1-minutes to allow workers to get setup and the deployment to get into a stable state
 	, TimeSinceLastCheckedTotalActorCounts(0.0f)
 	, bHasRequiredPlayersCheckFailed(false)
-	, RequiredPlayerCheckTimer(11*60) // 1-minute later then RequiredPlayerReportTimer to make sure all the workers had reported their migration
-	, DeploymentValidTimer(16*60) // 16-minute window to check between
+	, RequiredPlayerCheckTimer(11*60) // all clients should have joined by this point
```
simonsarginson, line 105: seconds / minutes? worth calling out?

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
simonsarginson: +1

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -481,6 +560,45 @@ void UTestGymsReplicationGraphNode_AlwaysRelevant_ForConnection::ResetGameWorldS
 	AlwaysRelevantStreamingLevelsNeedingReplication.Empty();
 }
 
+void UTestGymsReplicationGraphNode_AlwaysRelevant_ForConnection::GatherClientInterestedActors(const FConnectionGatherActorListParameters& Params)
+{
+	InterestedActorList.Reset();
```
simonsarginson, line 565: `InterestedActorList.Reset(Params.Viewers.Num()*3);`
Take a greedy guess at space requirements and reserve?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -481,6 +560,45 @@ void UTestGymsReplicationGraphNode_AlwaysRelevant_ForConnection::ResetGameWorldS
...
+			}
+		}
+	}
+
+	Params.OutGatheredReplicationLists.AddReplicationActorList(InterestedActorList);
+
+	UTestGymsReplicationGraph* TestGymsGraph = CastChecked<UTestGymsReplicationGraph>(GetOuter());
+	TMap<FName, FActorRepListRefView>& AlwaysRelevantStreamingLevelActors = TestGymsGraph->AlwaysRelevantStreamingLevelActors;
+
+	for (int32 Idx = AlwaysRelevantStreamingLevelsNeedingReplication.Num() - 1; Idx >= 0; --Idx)
```
simonsarginson, line 584: Why backwards?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -749,6 +880,35 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::GatherActorListsForConnecti
 	Params.OutGatheredReplicationLists.AddReplicationActorList(ReplicationActorList);
 }
 
+void UTestGymsReplicationGraphNode_GlobalViewTarget::GatherClientInterestedActors(const FConnectionGatherActorListParameters& Params)
+{
+	ReplicationActorList.Reset();
```
simonsarginson, line 885: reserve size?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -757,6 +917,87 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::LogNode(FReplicationGraphDe
...
+void UTestGymsReplicationGraphNode_NearestActors::NotifyAddNetworkActor(const FNewReplicatedActorInfo& ActorInfo)
+{
+	ReplicationActorList.Add(ActorInfo.Actor);
+}
+
+bool UTestGymsReplicationGraphNode_NearestActors::NotifyRemoveNetworkActor(const FNewReplicatedActorInfo& ActorInfo, bool bWarnIfNotFound /*= true*/)
+{
+	bool bRemovedSomething = false;
+	bRemovedSomething = ReplicationActorList.RemoveFast(ActorInfo.Actor);
+	return bRemovedSomething;
```
simonsarginson, line 930:
```
	return ReplicationActorList.RemoveFast(ActorInfo.Actor);
```
Could just be this?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -757,6 +917,87 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::LogNode(FReplicationGraphDe
...
+
+void UTestGymsReplicationGraphNode_NearestActors::GatherActorListsForConnection(const FConnectionGatherActorListParameters& Params)
+{
+	// Return all actors for replication
+	if (ReplicationActorList.Num() > 0)
+	{
+		FGlobalActorReplicationInfoMap& GlobalMap = *GraphGlobals->GlobalActorReplicationInfoMap;
+
+		// Cache actor location
+		for (FActorRepListType& Actor : ReplicationActorList)
```
simonsarginson, line 941: const?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -757,6 +917,87 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::LogNode(FReplicationGraphDe
...
+		SortedActors.Reset(ActorCount);
+
+		ensure(Params.Viewers.Num() == 1);	// Don't support multiple viewers for interest calculation
+		const FNetViewer& Viewer = Params.Viewers[0];
+
+		for (AActor* Actor : ReplicationActorList)
+		{
+			FGlobalActorReplicationInfo& ActorRepInfo = GlobalMap.Get(Actor);
+
+			float DistanceToViewer = (Viewer.ViewLocation - ActorRepInfo.WorldLocation).SizeSquared();	// check for max size?
```
simonsarginson, line 970: const?

Game/Source/GDKTestGyms/Private/UserExperienceReporter.cpp
```diff
@@ -100,7 +100,7 @@ void UUserExperienceReporter::ReportMetrics()
 				bFrameRateValid = false;
 				if (const USpatialNetDriver* NetDriver = Cast<USpatialNetDriver>(GameInstance->GetWorld()->GetNetDriver()))
 				{
-					UE_LOG(LogTemp, Error, TEXT("Client %s failed with fps %f below min threshold %f"), *NetDriver->Connection->GetWorkerId(), GameInstance->GetAveragedFPS(), Constants->GetMinClientFPS());
```
simonsarginson: LogTemp?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
```diff
@@ -855,12 +855,12 @@ void UTestGymsReplicationGraphNode_GlobalViewTarget::GatherActorListsForConnecti
 
 	for (const FNetViewer& CurViewer : Params.Viewers)
 	{
-		if (CurViewer.InViewer)
+		if (CurViewer.InViewer != nullptr)
```
simonsarginson, line 858:
```
if (IsValid(CurViewer.InViewer))
```

Game/Source/GDKTestGyms/Private/UserExperienceComponent.cpp
m-samiec: Good point - will unbind on destroy

Game/Source/GDKTestGyms/Private/UserExperienceReporter.cpp
m-samiec: Yeah - we use it here and elsewhere in small classes within the test gyms

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: Yeah it would be invalid to pass it anything but a number. We also output the values set by these calls in a general log at the bottom of this function, so the user could debug if it's a problem.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: This runs on non-auth servers now. Part of the "allow UX reporting to work in zoning scenarios commit".

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
simonsarginson: But it won't be an error, could be difficult to spot if we are just leaning on printing output values?

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
m-samiec: This will typically be the same from frame to frame, and reset will allow the internal array to persist its size, so don't think setting any size is necessary.

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
m-samiec: Based on existing code in GatherActorListsForConnection.

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
m-samiec: Will persist size, so doesn't seem necessary.

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
m-samiec: Have seen that ConditionalAdd actually does validation checks internally, so have removed this check entirely.

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
simonsarginson: it doesn't shrink on reset, fair dues.

Game/Source/GDKTestGyms/Private/TestGymsReplicationGraph.cpp
simonsarginson: Seems a bit awkward to ignore part of the API surface of this virtual function though? I would be pretty surprised if I called this on a base class pointer and it did not warn.

m-samiec: Oh you guys are right, I completely missed the `bWarnIfNotFound` variable - will fix.

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
m-samiec: I think this is fine. If there was an issue with the scenario, first thing you'd check is the settings.

simonsarginson: ok agree to disagree, not a fan of silent errors, you have more context so consider it resolved : ).

simonsarginson (approved):
LGTM

Game/Source/GDKTestGyms/BenchmarkGymGameModeBase.cpp
MatthewSandfordImprobable: Perhaps AveragedClientRTTMS should be WorstAveragedClientRTTMS

Game/Source/GDKTestGyms/Private/UserExperienceReporter.cpp
```diff
@@ -96,6 +97,10 @@ void UUserExperienceReporter::ReportMetrics()
 			if (Constants->ClientFPSMetricDelay.HasTimerGoneOff() && GameInstance->GetAveragedFPS() < Constants->GetMinClientFPS())
 			{
 				bFrameRateValid = false;
+				if (const USpatialNetDriver* NetDriver = Cast<USpatialNetDriver>(GameInstance->GetWorld()->GetNetDriver()))
```
MatthewSandfordImprobable, line 100: Probably want some log even if netdriver is null. Maybe the worker id can just be invalid if null driver.

MatthewSandfordImprobable (approved):

PR #478, opened 2021-11-24
[INT-57] Zero-initialize UProperties to remove some warnings

mironec:
Also https://github.com/spatialos/UnrealGDK/pull/3342


simonsarginson (approved):

Game/Source/GDKTestGyms/Public/BlackboardValues.h
```diff
@@ -12,11 +12,11 @@ struct GDKTESTGYMS_API FBlackboardValues
 	GENERATED_BODY()
 
 	UPROPERTY()
-	FVector TargetAValue; // Points to run between
+	FVector TargetAValue = FVector(EForceInit::ForceInit); // Points to run between
```
m-samiec, line 15: Haven't seen EForceInit::ForceInit before - why is it required?

mironec: So I found this while looking at the FVector constructors. It can either take nothing (being uninitialized), xyz, or this force init enum, which can be ForceInit or ForceInitToZero. So I just chose ForceInit, since it didn't matter to me what it was, but maybe it should be ForceInitToZero?

simonsarginson: re: original question iirc UE now warns (errors?) if values are not initialised right @mironec ?

mironec: ah, yes

m-samiec: If we have to pick one, might as well ForceToZero, it's not like the performance trade-offs will be noticeable in this instance. Is there a benefit in using ForceToZero over FVector::Zero?

mironec: whooooops
I merged

mironec: in the fvector case, pretty sure FVector::Zero and ForceInitToZero is equivalent
however, EForceInit is accepted in constructors elsewhere as well, such as color, box, etc.

